PR #5, opened 2016-04-28
Initial code import

### LINKED ISSUE #3: Enable Google code style checking
See: GoogleCloudPlatform/app-tools-lib-for-java#17

meltsufin:
Also fixes #3.


src/main/java/com/google/cloud/tools/maven/DeployMojo.java
```diff
@@ -0,0 +1,163 @@
...
+@Execute(phase = LifecyclePhase.PACKAGE)
+public class DeployMojo extends StageMojo implements DeployConfiguration {
+
+  /**
+   * The yaml files for the services or configurations you want to deploy. If not given, defaults to
+   * app.yaml in the current directory. If that is not found, attempts to automatically generate
+   * necessary configuration files (such as app.yaml) in the current directory.
+   */
+  @Parameter(name = "deployDeployables", property = "app.deploy.deployables")
+  private List<File> deployables;
```
meltsufin: Is this what we're thinking in terms of the parameter names and command line property names? The convention looks a little funny for some of the fields. For example, 'deployDeplyables or 'stageStagingDirectory'. Is this OK? @loosebazooka @patflynn

patflynn: is there no way to configure the xml element name to be something other than the field name?  I was initially picturing the namespacing being punctuated somehow. e.g. stage.stagingDirectory

meltsufin: The name attribute of the @Parameter sets the XML element name. We can use the periods there.

patflynn: in that case there's no need to prefix the actual fields on this class IMO

pom.xml
```diff
@@ -0,0 +1,116 @@
...
+      <artifactId>maven-core</artifactId>
+      <version>3.3.9</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.tools</groupId>
+      <artifactId>app-tools-lib-for-java</artifactId>
+      <version>0.1.0-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.aether</groupId>
```
loosebazooka: do we really need this dependency here?

pom.xml
```diff
@@ -0,0 +1,116 @@
...
+      <artifactId>aether-api</artifactId>
+      <version>1.0.2.v20150114</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
```
loosebazooka: what are we using commons.io for?

src/main/java/com/google/cloud/tools/maven/DeployMojo.java
```diff
@@ -0,0 +1,157 @@
...
+@Mojo(name = "deploy")
+@Execute(phase = LifecyclePhase.PACKAGE)
+public class DeployMojo extends StageMojo implements DeployConfiguration {
+
+  /**
+   * The yaml files for the services or configurations you want to deploy. If not given, defaults to
+   * app.yaml in the current directory. If that is not found, attempts to automatically generate
+   * necessary configuration files (such as app.yaml) in the current directory.
+   */
+  @Parameter(alias = "deploy.deployables", property = "app.deploy.deployables")
```
loosebazooka: alias vs name?

src/main/java/com/google/cloud/tools/maven/RunAsyncMojo.java
```diff
@@ -0,0 +1,26 @@
...
+public class RunAsyncMojo extends RunMojo {
+
+  private CountDownLatch waitStartedLatch;
+
+  /**
+   * Configures an asynchronous process runner for running server.
+   */
+  public RunAsyncMojo() {
+    super();
+    processRunner.setAsync(true);
```
loosebazooka: I think users needs to be allowed to define where the devappserver output goes when running in the background.

loosebazooka: or at least we should point it to a tmp file in `target`

pom.xml
meltsufin: We don't need it right now. It was used in the old plugins to find the App Engine SDK in the Maven repo dynamically. I'll remove it.

pom.xml
meltsufin: I'm using it in `StageMojo` to delete the staging directory `FileUtils.deleteDirectory(stagingDirectory)`. Guava doesn't have a method for that. I wonder if this logic should move to the common lib, though.

src/main/java/com/google/cloud/tools/maven/DeployMojo.java
meltsufin: It turns out `alias` is what needs to be used. `name` is used when you want to use a getter for the property. The documentation on this is not super clear, but I found this out by experimenting.

src/main/java/com/google/cloud/tools/maven/RunAsyncMojo.java
meltsufin: If we redirect the output to the parent process (Maven), they can redirect it wherever they want. What do you think?

loosebazooka: I think with redirectIO we wont be able to detect that the server is started, I did something like this : https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/4/files#diff-3bf6c770f2ac694a154614998f389123R60

loosebazooka:
Is the plan to alias the params in all Mojos?

meltsufin:
Yes, I'm going to alias all parameters that are not the same as the field name.

src/main/java/com/google/cloud/tools/maven/RunAsyncMojo.java
meltsufin: I want to experiment more with this, and whatever the solution is, we should move the common code to the lib.

pom.xml
```diff
@@ -0,0 +1,111 @@
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.cloud.tools.maven</groupId>
```
loosebazooka: I think we're using
`<groupId>com.google.cloud.tools</group>`

meltsufin: Yes makes sense!

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -47,22 +44,9 @@
...
-        throw new MojoFailureException(
-            "Cloud SDK Path was not provided, and cannot be automatically detected.", e);
-      }
-    }
-    if (processRunner == null) {
-      this.cloudSdk = new CloudSdk(cloudSdkPath);
-    } else {
-      this.cloudSdk = new CloudSdk(cloudSdkPath, processRunner);
-    }
+    this.cloudSdk = new CloudSdk.Builder()
```
loosebazooka: is this change in the common lib yet?

meltsufin: yep, very recently https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/54

loosebazooka:
:ship: Okay lets merge this.

meltsufin:
I have to update the readme

src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -0,0 +1,358 @@
...
+      }
+    };
+    processRunner.setStdOutLineListener(listener);
+  }
+
+  @Override
+  public void execute() throws MojoExecutionException, MojoFailureException {
+    super.execute();
+
+    waitStartedLatch = new CountDownLatch(1);
```
meltsufin: I would like to move this to the common lib somehow.

loosebazooka:
Sure, I think we're a good state now though. Lets put all subsequent changes in new PRs though, so much easier to review smaller scoped PRs?

meltsufin:
Sounds good.

loosebazooka:
:rocket:

PR #9, opened 2016-05-12
Use tools lib to handle waiting for dev app server to start

meltsufin:
The functionality to wait for an async process to start successfully has been moved to the common tools library. See: https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/67.

Also, changed pom.xml to generated sources jar and to run faster.

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
@@ -50,8 +50,8 @@ public abstract class CloudSdkMojo extends AbstractMojo {
 
     cloudSdkBuilder = new CloudSdk.Builder()
         .sdkPath(cloudSdkPath)
-        .stdOutLineListener(gcloudOutputListener)
-        .stdErrLineListener(gcloudOutputListener);
+        .addStdOutLineListener(gcloudOutputListener)
+        .addStdErrLineListener(gcloudOutputListener);
   }
 
   protected CloudSdk getCloudSdk() {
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/maven/RunAsyncMojo.java
```diff
@@ -29,14 +30,22 @@
...
-  }
+  @Parameter(defaultValue = "30",
+      alias = "devserver.startSuccessTimeout", property = "app.devserver.startSuccessTimeout")
+  protected int startSuccessTimeout;
 
+  @Override
+  public void execute() throws MojoExecutionException, MojoFailureException {
+    cloudSdkBuilder
+        .async(true)
+        .waitSuccessMessage("Dev App Server is now running")
```
meltsufin: This line should probably move to the lib.

meltsufin:
@loosebazooka PTAL

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -38,15 +36,25 @@
...
   protected File cloudSdkPath;
 
-  protected CloudSdk cloudSdk;
+  protected CloudSdk.Builder cloudSdkBuilder;
 
-  protected DefaultProcessRunner processRunner;
+  private final ProcessOutputLineListener gcloudOutputListener = new ProcessOutputLineListener() {
+    @Override
+    public void outputLine(String line) {
+      System.out.println("GCLOUD: " + line);
```
loosebazooka: use logger of some kind here?

meltsufin: the obvious choice is the Maven logger, but I still want to indicate that output is coming from GCLOUD somehow.

loosebazooka:
:ship:

Issue #4, opened 2016-04-26
Set an env variable to track server-side usage of gcloud


chanseokoh:
The same thing as in [here](https://github.com/GoogleCloudPlatform/app-gradle-plugin/issues/3).

meltsufin:
@chanseokoh has it been settled what we're actually passing in for CLOUDSDK_METRICS_ENVIRONMENT and CLOUDSDK_METRICS_ENVIRONMENT_VERSION?

chanseokoh:
For IntelliJ, we pass `gcloud-intellij`. For others, you may suggest and use something similar along the line. For VERSION, let's pass the version of the plugins. For IntelliJ, it would probably be like `0.9.5-beta`.

PR #16, opened 2016-05-20
Set env var and tag plugin identify for metrics

chanseokoh:
In relation to #4.

I'd like to set this now so that we don't forget it before the next release. How about "app-maven-plugin"?

Note: we still need to set ENVIRONMENT_VERSION.

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
@@ -50,6 +50,7 @@ public abstract class CloudSdkMojo extends AbstractMojo {
 
     cloudSdkBuilder = new CloudSdk.Builder()
         .sdkPath(cloudSdkPath)
+        .appCommandMetricsEnvironment("app-maven-plugin")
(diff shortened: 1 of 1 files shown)


meltsufin:
Why not set the ENVIRONMENT_VERSION as part of this PR?

chanseokoh:
I can do that, but I'm not familiar with Maven. I see it's currently 0.1.0-SNAPSHOT in pom.xml. How can I retrieve it?

meltsufin:
I found a way to get this info. See my commit. I had to move this out of the constructor though because of the component injection.

chanseokoh:
Great! Shall we merge this?

meltsufin:
Sure :ship:

chanseokoh:
Just to confirm: `pluginDescriptor.getArtifactId()` will return "app-maven-plugin", won't it?

meltsufin:
Yes, it should.

chanseokoh:
Sorry. On second thought, isn't it "gcp-app-maven-plugin" if you look at pom.xml?

meltsufin:
Sorry, yes, missed that prefix too. :)

PR #18, opened 2016-05-20
Enable FindBugs and fix violations

### LINKED ISSUE #14: Enable FindBugs plugin and fix any violations

meltsufin:
Fixes #14


meltsufin:
@loosebazooka PTAL

pom.xml
```diff
@@ -47,35 +48,36 @@
...
-      <version>3.3.9</version>
-    </dependency>
+
     <dependency>
       <groupId>com.google.cloud.tools</groupId>
       <artifactId>app-tools-lib-for-java</artifactId>
       <version>0.1.0-SNAPSHOT</version>
     </dependency>
+
     <dependency>
```
loosebazooka: what is commons.io used for that guava can't do?

meltsufin: Guava doesn't provide a method to delete a directory recursively. It used to, but the new versions don't.

loosebazooka: Oh I see, to clean the staging directory.

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -47,20 +46,20 @@ public void outputLine(String line) {
     }
   };
 
-  @Component
+  @Parameter(defaultValue = "${pluginDescriptor}", readonly = true)
```
loosebazooka: Why is this a parameter?

pom.xml
meltsufin: Yep. Are you doing this in Gradle too?

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
meltsufin: It's the same thing. `@Component ` is now deprecated. Hence the change. `readonly=true` makes it invisible to users of the plugin.

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -169,10 +169,12 @@ public void execute() throws MojoExecutionException, MojoFailureException {
       try {
         FileUtils.deleteDirectory(stagingDirectory);
       } catch (IOException e) {
-        new MojoFailureException("Unable to delete staging directory.", e);
+        throw new MojoFailureException("Unable to delete staging directory.", e);
```
meltsufin: A bug that findbugs actually found!

pom.xml
loosebazooka: gradle implements delete themselves

loosebazooka:
:ship:

PR #21, opened 2016-05-25
Add unit tests

### LINKED ISSUE #2: Implement missing unit tests

meltsufin:
Fixes #2 and #20.

I've also lightly refactored the code to make testing easier.

@loosebazooka PTAL


src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -38,31 +38,27 @@
   @Parameter(property = "cloudSdkPath", required = false)
   protected File cloudSdkPath;
 
-  protected CloudSdk.Builder cloudSdkBuilder;
-
-  private final ProcessOutputLineListener gcloudOutputListener = new ProcessOutputLineListener() {
+  protected final ProcessOutputLineListener gcloudOutputListener = new ProcessOutputLineListener() {
```
loosebazooka: This feels like something that should go in the constructor? But that seems like a preference thing.

meltsufin: I don't really have a preference, but willing to change, if there is some best-practice that it's violating.

src/test/java/com/google/cloud/tools/maven/CloudSdkMojoTest.java
```diff
@@ -0,0 +1,80 @@
...
+    return pluginDescriptorMock;
+  }
+
+  public static void verifyCloudSdkCommon(CloudSdkMojo mojo, CloudSdk.Builder cloudSdkBuilderSpy) {
+    assertNotNull(mojo.gcloudOutputListener);
+    verify(cloudSdkBuilderSpy).sdkPath(mojo.cloudSdkPath);
+    verify(cloudSdkBuilderSpy).addStdOutLineListener(mojo.gcloudOutputListener);
+    verify(cloudSdkBuilderSpy).addStdErrLineListener(mojo.gcloudOutputListener);
+    verify(cloudSdkBuilderSpy).appCommandMetricsEnvironment("gcp-app-maven-plugin");
+    verify(cloudSdkBuilderSpy).appCommandMetricsEnvironmentVersion("0.1.0");
```
loosebazooka: I wonder if checking the version like this is the way to go? Seems like an update to the version would require the developer to update all the tests.

loosebazooka:
I think it looks okay, there seems to be a lot of use of mockito spy. To me, that feels like we could do a better job with the encapsulation/provisioning of functionality. I'm open to be convinced otherwise though for these tests.

src/test/java/com/google/cloud/tools/maven/CloudSdkMojoTest.java
meltsufin: It's a dummy version hard-coded in the test itself, but it's misleading because I should've used the constant here as well. I'll fix it.

meltsufin:
I'm not a fan of using spies either, but it seemed to make sense here. The spies are only used for the CloudSdk.Builder and the actual mojos that are under test. The reason for spying on CloudSdk.Builder is to verify that the mojo is configuring the CloudSdk correctly. You can also achieve this by mocking the CloudSdk.Builder, but it seems to add unnecessary boilerplate code.

The reason for wrapping the mojos in a spy is a little less obvious. It's mostly to be able to overwrite methods like the `createCloudSdkBuilder()` with mocks. However, I think this can be improved with better encapsulation like you're suggesting.

Let me see if I can do something to eliminate these spies.

meltsufin:
This may help with mocking the CloudSdk.Builder: http://blogs.atlassian.com/2013/02/mockito-makes-mocking-fluid-interfaces-easy/

meltsufin:
@loosebazooka PTAL

I got rid of all the spies. I'm using RETURNS_SELF for the CloudSdk.Builder mocks; and I created a Factory for instantiating depedencies, which allowed me to get rid of the spies on the actual mojos under test.

src/main/java/com/google/cloud/tools/maven/Factory.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.cloud.tools.app.impl.cloudsdk.internal.process.ProcessOutputLineListener;
+import com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk;
+
+import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.logging.Log;
+
+/**
+ * Factory for dependencies.
+ */
+public class Factory {
```
loosebazooka: Is Factory too generic a name?

meltsufin: How about DependenciesFactory?

loosebazooka: It seems like a cloudsdk specific factory. Perhaps something that indicates that.

src/main/java/com/google/cloud/tools/maven/Factory.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public AppEngineDeployment deployment(CloudSdk cloudSdk) {
+    return new CloudSdkAppEngineDeployment(cloudSdk);
+  }
+
+  public AppEngineDevServer devServer(CloudSdk cloudSdk) {
+    return new CloudSdkAppEngineDevServer(cloudSdk);
+  }
+
+  public AppEngineDevServer devServer() {
```
loosebazooka: I know that this is for stopping only. Is there some way to indicate that?

src/main/java/com/google/cloud/tools/maven/Factory.java
meltsufin: The return types are all non-cloudsdk-specific interfaces. So, the idea is that we can swap out the implementations later. What do you think of `AppEngineDependenciesFactory`?

src/main/java/com/google/cloud/tools/maven/Factory.java
meltsufin: Sure, I'll just call it `devServerStop()`.

src/main/java/com/google/cloud/tools/maven/Factory.java
meltsufin: Actually there's also the Maven Log there, which has nothing to do with App Engine. I think I'll call it `PluginDependenciesFactory`.

src/main/java/com/google/cloud/tools/maven/PluginDependenciesFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public AppEngineDevServer devServer(CloudSdk cloudSdk) {
+    return new CloudSdkAppEngineDevServer(cloudSdk);
+  }
+
+  public AppEngineDevServer devServerStop() {
+    return new CloudSdkAppEngineDevServer(null);
+  }
+
+  public Log logger(AbstractMojo mojo) {
```
loosebazooka: Perhaps this makes the scope too broad. It really is just producing AppEngine actions. Isn't the logging framework usually responsible for handling the logging actions? Maybe I don't understand why this is helping us

meltsufin: There isn't really a logging framework here. If we want to monitor the log for testing, we need to use `@Spy` to spy on `AbstractMojo.getLog()`. Let me see if I can work around this.

meltsufin:
@loosebazooka I found a way to inject a mock logger, while still using the Maven convention of getLog(), without involving the factory in it. The factory is now called AppEngineDependenciesFactory. What do you think?

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -36,31 +35,19 @@
...
-        .addStdOutLineListener(gcloudOutputListener)
-        .addStdErrLineListener(gcloudOutputListener);
-  }
+  protected AppEngineDependenciesFactory factory = new AppEngineDependenciesFactory();
 
-  protected CloudSdk getCloudSdk() {
+  protected CloudSdk.Builder configureCloudSdkBuilder(CloudSdk.Builder cloudSdkBuilder) {
     return cloudSdkBuilder
         .sdkPath(cloudSdkPath)
+        .addStdOutLineListener(factory.gcloudOutputListener(this))
```
loosebazooka: The factory is being used cyclically here isn't it?

sdk (factory)
factory.deploy(sdk) ?

meltsufin:
@loosebazooka  Before I update the tests, can you take a look at this refactoring to see if it's on the right track?

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -0,0 +1,130 @@
...
+      this.log = log;
+    }
+
+    @Override
+    public void outputLine(String line) {
+      log.info("GCLOUD: " + line);
+    }
+  }
+
+  protected static class CloudSdkFactory {
```
loosebazooka: I feel like this class is not necessary. Does this help with testing somehow or something that isn't apparent yet?

meltsufin: It helps with testing. It will be more apparent when you see the tests.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -0,0 +1,130 @@
...
+    return cloudSdkFactory.flexibleStaging();
+  }
+
+  @Override
+  public AppEngineDeployment deployment() {
+    return cloudSdkFactory.deployment(defaultCloudSdkBuilder().build());
+  }
+
+  @Override
+  public AppEngineDevServer devServerRunSync() {
```
loosebazooka: does this need to go into the cloudSdkFactory then?

meltsufin: I'm trying to keep the CloudSdkFactory as dumb as possible, because we won't be able to test it.

loosebazooka:
anyway, that minor thing aside, I like this much better.

meltsufin:
Cool, I'll fix up the tests and give you another look.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
loosebazooka: I mean, should this be return cloudSdkFactory.devServer(defaultCloudSdkBuilder().build())?

meltsufin: Oh! good catch!

meltsufin:
I fixed up the unit tests to work with the refactored code.
@loosebazooka PTAL

loosebazooka:
I think this is good, looks like most of the logic we want to test is in the common lib anyway.

:Ship:

loosebazooka:
:ship:

meltsufin:
Thanks!

PR #32, opened 2016-06-08
sync with common library package reorg

meltsufin:
@loosebazooka PTAL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ cache:
   directories:
    - $HOME/.m2
 
+# skip install
+install: true
+
 script: "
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -6,6 +6,9 @@ cache:
   directories:
    - $HOME/.m2
 
+# skip install
```
loosebazooka: kinda confusing? `skip install` and `install: true`

loosebazooka:
small comment otherwise :ship:

.travis.yml
meltsufin: I know, hence the comment, but that's what Travis docs recommend: https://docs.travis-ci.com/user/customizing-the-build/#Skipping-the-Installation-Step

PR #36, opened 2016-06-13
Add Integration Tests

### LINKED ISSUE #34: Add Integration Tests

meltsufin:
Fixes #34.


src/test/java/com/google/cloud/tools/maven/it/DeployMojoIT.java
```diff
@@ -0,0 +1,65 @@
...
+package com.google.cloud.tools.maven.it;
+
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.apache.maven.it.util.ResourceExtractor;
+import org.junit.Test;
+
+import java.io.IOException;
+
+public class DeployMojoIT extends AbstractMojoIT {
```
patflynn: Even though the 'I' and 'T' are for separate words we still case them 'It'. 

Think HttpRequest instead of HTTPRequest.  It's prettier that way.

meltsufin: Interesting... I'll update.

patflynn:
actually, come to think of it, doesn't Maven usually have you put integration tests under src/integration-test/..?  Also perusing around for Maven integration test conventions I'm seeing ..IT as an actual thing, although I'm not a big fan and it seems to go against standard Java class naming.  Part of me wonders if the src is siloed to integration-test whether we need to further qualify them.  Internally we often just call them ...Test.

meltsufin:
I'm not aware of the `src/integration-test` as a Maven convention for integration tests, but it can definitely be configured that way.
The failsafe plugin that is usually used to run integration tests in Maven has the default of looking for `**/*IT.java`. I'm OK with using `It` or even `Test`, although I've always used the `IT` suffix before for integration tests.

.travis.yml
```diff
@@ -6,8 +6,21 @@ cache:
   directories:
    - $HOME/.m2
 
-# skip install
-install: true
+install:
+  # download Cloud SDK
+  - wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-112.0.0-linux-x86_64.tar.gz
```
loosebazooka: I think you can get "latest" too if you wanted to.

meltsufin: Do you have a URL for "latest'?

src/test/java/com/google/cloud/tools/maven/it/DeployMojoIt.java
```diff
@@ -0,0 +1,82 @@
...
+    verifier.verifyErrorFreeLog();
+    verifier.verifyTextInLog("Detected App Engine flexible environment application");
+    verifier.verifyTextInLog("GCLOUD: Deployed service");
+
+    // cleanup
+    deleteService("flexible-project");
+  }
+
+  private void deleteService(String service) throws ProcessRunnerException {
+    CloudSdk cloudSdk = new CloudSdk.Builder().build();
```
loosebazooka: maybe add an excepting exit listener here?

.travis.yml
```diff
@@ -6,8 +6,21 @@ cache:
   directories:
    - $HOME/.m2
 
-# skip install
-install: true
+install:
```
loosebazooka: can we cache some of these steps? Or is the download install update pretty fast?

.travis.yml
```diff
@@ -6,8 +6,21 @@ cache:
...
+  # download Cloud SDK
+  - wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-112.0.0-linux-x86_64.tar.gz
+  - tar -xzvf google-cloud-sdk-112.0.0-linux-x86_64.tar.gz
+  - export PATH=$PWD/google-cloud-sdk/bin:$PATH
```
loosebazooka: can we also export $CLOUDSDK_CORE_DISABLE_USAGE_REPORTING=true so we don't mess with metrics reporting?

.travis.yml
meltsufin: I think you're right about caching. The part that takes the longest is the app engine component install. ~30 seconds.

.travis.yml
meltsufin: Will do

.travis.yml
meltsufin: I'm not convinced caching will improve it much because it is actually a lot of data. I'm going to leave this for a future build speed optimization task.

src/test/java/com/google/cloud/tools/maven/it/StopMojoIt.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // stop dev app server
+    verifier.setLogFileName("testStopStandard.txt");
+    verifier.setAutoclean(false);
+    verifier.executeGoal("gcp-app:stop");
+
+    // verify dev app server is down
+    verifier.verifyErrorFreeLog();
+    // wait up to 5 seconds for the server to stop
+    for (int i = 0; i < 50 && getUrlContent(SERVER_URL) != null; i++) {
```
loosebazooka: Not sure I'm convinced of the style here. Why not just make the test wait 5 seconds?

loosebazooka: Actually, I can kinda see how it would make sense. The assertion after checking the same condition in the forloop just feels weird.

meltsufin: I'll work on making this whole mechanism clearer

meltsufin:
@patflynn do you want to stick with `*It` for integration tests, or switch to `*IT`?

patflynn:
Do whatever the convention is. Which I think is IT.

loosebazooka:
any idea why appveroy is failing?

meltsufin:
I'm trying to figure out the right syntax with yaml and CMD for AppVeyor. Getting close...

loosebazooka:
Lets drop the appveyor for integration tests for now.

.travis.yml
```diff
@@ -6,15 +6,29 @@ cache:
...
+  - export CLOUDSDK_CORE_DISABLE_USAGE_REPORTING=true
+  # update all Cloud SDK components
+  - gcloud components update --quiet
+  # add App Eengine component to Cloud SDK
+  - gcloud components install app-engine-java --quiet
```
loosebazooka: can you file a bug against yourself to track or point to forum request at appveyor for this google-cloud-sdk install they might have on the image?

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -33,7 +33,7 @@
   @Parameter(property = "cloudSdkPath", required = false)
   private File cloudSdkPath;
 
-  @Parameter(defaultValue = "${pluginDescriptor}", readonly = true)
+  @Parameter(defaultValue = "${plugin}", readonly = true)
```
loosebazooka: Just for information : why this change?

loosebazooka:
any idea why the devappserver test is failing? otherwise lgtm

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
meltsufin: I turned out that ${pluginDescriptor} doesn't work, but ${plugin} does.

meltsufin:
The synchronous dev appserver test is flaky for some reason. I'm going to look at it a bit more this morning.

PR #37, opened 2016-06-15
User docker directory instead of Dockerfile for flex

meltsufin:
Syncs up with the common library changes.


meltsufin:
@loosebazooka PTAL

loosebazooka:
:shipit: :rocket: :blue_car:

PR #41, opened 2016-06-20
specify appengine config directory rather than app.yaml path

meltsufin:
This is a sync up with the common tools library change.
@loosebazooka


loosebazooka:
:ship:

patflynn:
This is a bit different from the way that gcloud models it.  They derive the directory context from the location of the app.yaml parameter.

meltsufin:
We can do it either way. It's even conceivable to overload the parameter to accept either app.yaml or directory path. @loosebazooka @arouzrokh

loosebazooka:
I think the problem is that gcloud doesn't try to deploy all the yaml files along with it. In gradle, I'm accepting both, the library code switches based on whether it's a single directory or a list of files. Even considering all that, the solution I put in might need to be worked on.

PR #45, opened 2016-06-24
Add pom.xml metadata, javadoc, and signing

meltsufin:
Added more metadata that is required for a release.
Also, added javadoc and artifact signing in a separate 'release' profile.
@patflynn


patflynn:
LGTM

PR #47, opened 2016-06-28
Improve integration tests

meltsufin:
* Stream maven output to console for integration tests
* Make Dev App Server tests more robust

### DIFF
--- a/src/test/java/com/google/cloud/tools/maven/it/AbstractMojoIntegrationTest.java
+++ b/src/test/java/com/google/cloud/tools/maven/it/AbstractMojoIntegrationTest.java
@@ -36,7 +36,7 @@ public abstract class AbstractMojoIntegrationTest {
 
   private static boolean doneInstallPlugin = false;
 
-//  @BeforeClass
+  @BeforeClass
   public static void installPlugin() throws VerificationException {
     // install the plugin under test
     if (!doneInstallPlugin) {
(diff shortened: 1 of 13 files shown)


meltsufin:
@loosebazooka PTAL
I made some improvements to the tests and how they execute on Travis and AppVeyor.

src/test/java/com/google/cloud/tools/maven/it/AbstractMojoIntegrationTest.java
```diff
@@ -31,13 +36,11 @@
 
   private static boolean doneInstallPlugin = false;
 
-  @BeforeClass
+//  @BeforeClass
```
loosebazooka: comment introduces some uncertainty, do we want a beforeclass here or no?

meltsufin: Great catch! I did this temporarily during testing and forgot to change it back.

src/test/java/com/google/cloud/tools/maven/it/AbstractMojoIntegrationTest.java
```diff
@@ -104,4 +107,53 @@ private static String getUrlContentWithRetries(String url, long timeoutMs, long
 
     return content;
   }
+
+  protected Verifier createStandardVerifier(String testName)
```
loosebazooka: I understand this is programming style thing, but would prefer these utilities be in their own classes and called into from the actual tests themselves.

meltsufin: I agree, I'll rework it.

loosebazooka:
:ship:

PR #53, opened 2016-07-11
Fix staging issue for compat Java 8 projects

### LINKED ISSUE #52: Standard staging fails for Compat Java 8 projects without Dockerfile

meltsufin:
Fixes #52 
Move up to version 0.1.3-SNAPSHOT of the appengine-plugins-core library

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/StageMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/StageMojo.java
@@ -197,7 +197,7 @@ public class StageMojo extends CloudSdkMojo implements StageStandardConfiguratio
       getLog().info("Detected App Engine standard environment application.");
 
       // force runtime to 'java' for compat projects using Java version >1.7
-      if (Float.parseFloat(getCompileTargetVersion()) > 1.7) {
+      if (Float.parseFloat(getCompileTargetVersion()) > 1.7f) {
         runtime = "java";
       }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -193,6 +196,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     if (new File(sourceDirectory.toString() + "/WEB-INF/appengine-web.xml").exists()) {
       getLog().info("Detected App Engine standard environment application.");
 
+      // force runtime to 'java' for compat projects using Java version >1.7
+      if (Float.parseFloat(getCompileTargetVersion()) > 1.7f) {
```
loosebazooka: do you have access to maven comparable version here?
https://maven.apache.org/ref/3.3.3/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html

meltsufin: I do, but it seems more verbose. I don't think the Java target version can ever be anything but a real number.

loosebazooka: alright

loosebazooka:
:rocket:

PR #56, opened 2016-07-14
Restrict Java 8 runtime forcing to compat projects

meltsufin:


loosebazooka:
appveyor change needs to go in first?

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -297,4 +309,45 @@ public File getDockerDirectory() {
   public String getRuntime() {
     return runtime;
   }
+
+
+  /**
+   * Simple parser for appengine-web.xml, this should ideally not exist, but we need it to correctly
+   * error when vm=false and the user is using java8 as the target platform
+   */
+  private static class AppEngineWebXml {
```
loosebazooka: I guess this is choice, but would have been nice to have this non-nested-class?

loosebazooka:
:runner:

src/main/java/com/google/cloud/tools/maven/StageMojo.java
meltsufin: Yes, but ultimately, if this is going to stay around, I would like it to be in the lib.

loosebazooka: good point

PR #57, opened 2016-07-15
Sync with common lib File -> Path change

meltsufin:


loosebazooka:
ugh really wish we could just use path everywhere...

:runner:

PR #58, opened 2016-07-18
Fail early when trying to start dev server with flexible app

### LINKED ISSUE #51: Provide better error messages when invoking Dev Appserver for non-standard env apps

meltsufin:
Fixes #51 

@loosebazooka

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/RunMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/RunMojo.java
@@ -221,8 +221,8 @@ public class RunMojo extends CloudSdkMojo implements RunConfiguration {
     return mavenProject != null
         && mavenProject.getBuild() != null
         && new File(
-        mavenProject.getBuild().getDirectory() + "/" + mavenProject.getBuild().getFinalName()
-            + "/WEB-INF/appengine-web.xml").exists();
+            mavenProject.getBuild().getDirectory() + "/"  + mavenProject.getBuild().getFinalName()
+                + "/WEB-INF/appengine-web.xml").exists();
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -214,6 +214,28 @@ protected void workAroundNonJava7Version() {
...
   }
 
+  /**
+   * Determine if the application is a Standard Environment app.
+   */
+  protected boolean isStandardEnvironmentApp() {
+    return mavenProject != null
+        && mavenProject.getBuild() != null
+        && new File(
+        mavenProject.getBuild().getDirectory() + "/" + mavenProject.getBuild().getFinalName()
```
loosebazooka: you want to check on the build directory? or the source directory?

loosebazooka: oh this is for the devappserver? Would it make sense to indicate the intention of the method to check the build directory to make it more explicit?

meltsufin: Since, we're running the Dev Appserver from the build dir, I thought it would be more appropriate to check there, rather than in the source. I'll modify the comment to make it more clear.

meltsufin:
@loosebazooka See comment updates.

meltsufin:
@loosebazooka PTAL

src/main/java/com/google/cloud/tools/maven/RunMojo.java
loosebazooka: feel like spacing is off here
```
&& new File(
    mavenProejct....
```

loosebazooka:
just a little thing, otherwise LGTM

PR #62, opened 2016-07-26
use travis_wait 30 to increase no output timeout to 30 mins

meltsufin:
The default timeout is 10 minutes, but the deployment test can sometimes take longer without producing any output to the console.

@loosebazooka


loosebazooka:
:shipit:

PR #63, opened 2016-08-05
Update to Plugins Core 0.1.5-SNAPSHOT and disable Travis remote tests

meltsufin:
The Travis deploy test is still unreliable. Removing it until the issue is resolved.

@loosebazooka PTAL


.travis.yml
```diff
@@ -37,7 +37,7 @@ install:
 script:
   - |
     if $INCLUDE_REMOTE_TESTS; then
```
loosebazooka: this if statements seems to have lost its meaning?

meltsufin: hmm yeah the variable name is not great

loosebazooka:
:ship:

PR #72, opened 2016-08-26
Remove application and version from appengine-web.xml test app

### LINKED ISSUE #67: test cases include the application id and version in the appengine-web.xml
storing the ID and version in the appengine-web.xml is no longer the recommended way to do it, and technically shouldn't work against the gcloud tooling.

meltsufin:
Fixes #67.

### DIFF
--- a/src/test/resources/projects/standard-project/pom.xml
+++ b/src/test/resources/projects/standard-project/pom.xml
@@ -22,6 +22,7 @@
         <artifactId>appengine-maven-plugin</artifactId>
         <version>${appengine-maven-plugin.version}</version>
         <configuration>
+          <version>1</version>
         </configuration>
       </plugin>
     </plugins>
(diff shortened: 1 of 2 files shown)


meltsufin:
Fails for now with `Bad configuration: No app id supplied and XML files have no <application> element`.
Waiting for the fix in AppCfg staging to make it into the Cloud SDK.

patflynn:
It sounds like we need to make the project ID flag required on the plugin
for now instead.  It's problematic that the staging code requires it to be
set in the xml if it's not passed as a flag, but strips it from the
generated app.yaml regardless.

loosebazooka:
What do we want to do about this PR here?

meltsufin:
@loosebazooka This is ready to be merged in now. It just needs your approval.

PR #79, opened 2016-09-06
Enable codecov-io for code coverage reporting in GitHub

meltsufin:


.travis.yml
```diff
@@ -39,6 +39,10 @@ script:
       mvn clean install -B -U
     fi
 
+# enable codecov-io
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
patflynn: this seems.. unsafe.  Is there no direct integration we can provide? Does it have to be through Travis?

meltsufin: This seems to be the recommended way to integrate it. We do have to trust them just like we trust Travis.

meltsufin: I will enable it only for pull requests that don't expose secrets.

meltsufin: Actually, it's not that straightforward. Pull requests from committers have access to secrets. I need to investigate this a bit more to find a workable solution.

meltsufin:
@patflynn Can you take a look and see if protecting the secrets the way I did is good?

.travis.yml
```diff
@@ -39,6 +39,16 @@ script:
       mvn clean install -B -U
     fi
 
+after_success:
+  # hide sensitive data from codecov-io
+  - rm -f config/travis-app-maven-plugin-e1bb638f290e.json
+  - export SONATYPE_USERNAME=[secure]
```
patflynn: why '[secure]'? I know that's the string that travis prints out when it assigns a secret value, but in our case we want to wipe it out.

meltsufin: You're right, doing `unset` on them is cleaner.

patflynn:
LGTM

PR #88, opened 2016-09-21
Maven now included in AppVeyor CI

meltsufin:
Removing installation of Maven in AppVeyor config, since it's now included in the default image.

@loosebazooka PTAL


meltsufin:
I can't find any info on it, but I think it's just latest stable, which works for us.

loosebazooka:
Lgtm

PR #97, opened 2016-09-30
Adds a genRepoInfoFile goal.

### LINKED ISSUE #100: Create source context during package creation

joaoandremartins:
Fixes #100.

This should be ready for early preview. I think I still need to add unit tests here.

For the generation to be done automatically, this has the sad shortcoming that the following needs to be added to the pom.xml of the project using the plugin:

      <plugin>
        <version>1.0.0-SNAPSHOT</version>

>         <executions>
>           <execution>
>             <phase>prepare-package</phase>
>             <goals>
>               <goal>genRepoInfoFile</goal>
>             </goals>
>           </execution>
>         </executions>

      </plugin>

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
@@ -28,6 +28,7 @@ import org.apache.maven.plugins.annotations.Parameter;
 
 import java.io.File;
 import java.nio.file.Paths;
+import java.util.logging.Logger;
 
 /**
  * Generates repository information files for the Stackdriver Debugger.
@@ -35,6 +36,11 @@ import java.nio.file.Paths;
 @Mojo(name = "genRepoInfoFile")
 @Execute(phase = LifecyclePhase.PREPARE_PACKAGE)
 public class GenRepoInfoFileMojo extends CloudSdkMojo implements GenRepoInfoFileConfiguration {
+
+  private static Logger logger = Logger.getLogger(GenRepoInfoFileMojo.class.getName());
+  private static String HOW_TO_FIX_MSG = "An error occurred while generating source context files."
+      + " To ignore source context generation errors, use the "
+      + "-Dapp.genRepoInfoFile.ignoreErrors=true flag.";
 
   /**
    * The root directory containing the source code of the app. Expected to be contained in a git
@@ -66,9 +72,9 @@ public class GenRepoInfoFileMojo extends CloudSdkMojo implements GenRepoInfoFile
       getAppEngineFactory().genRepoInfoFile().generate(this);
     } catch (AppEngineException aee) {
       if (!ignoreErrors) {
-        throw new MojoExecutionException("An error occurred while generating source context files."
-            + " To ignore source context generation errors, use the "
-            + "-Dapp.genRepoInfoFile.ignoreErrors=false flag.", aee);
+        throw new MojoExecutionException(HOW_TO_FIX_MSG, aee);
+      } else {
+        logger.warning(HOW_TO_FIX_MSG);
       }
     }
   }
--- a/src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
@@ -30,7 +30,7 @@ import java.io.File;
 import java.nio.file.Paths;
 
 /**
- * Generates source context files.
+ * Generates repository information files for the Stackdriver Debugger.
  */
 @Mojo(name = "genRepoInfoFile")
 @Execute(phase = LifecyclePhase.PREPARE_PACKAGE)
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Parameter(defaultValue = "${project.build.outputDirectory}", property = "outputDirectory")
+  private String outputDirectory;
+
+  @Parameter(defaultValue = "false", property = "ignoreSrcCtxError")
+  private boolean ignoreErrors;
+
+  @Override
+  public void execute() throws MojoExecutionException, MojoFailureException {
+    DefaultGenRepoInfoFileConfiguration genConfig = new DefaultGenRepoInfoFileConfiguration();
```
joaoandremartins: Done.

src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
```diff
@@ -0,0 +1,64 @@
...
+import java.nio.file.Paths;
+
+/**
+ * Generates source context files.
+ */
+@Mojo(name = "genRepoInfoFile")
+@Execute(phase = LifecyclePhase.PREPARE_PACKAGE)
+public class GenRepoInfoFileMojo extends CloudSdkMojo {
+
+  @Parameter(defaultValue = "${project.basedir}")
```
joaoandremartins: Done.

joaoandremartins:
@meltsufin PTAL

Now with integration and unit tests.

Also, I tried mucking around with creating a dummy git repo in standard-project by copying the .git folder of a dummy project to the standard-project source's root for the integration tests, but gen-repo-info-files didn't like it very much.

The way gen-repo-info-files works is to run actual git commands to populate the generated source context files, and the ones I ran used a metadata mix of the app-maven-plugin and the dummy git repos. (e.g., git status assumed I was adding the dummy app's files to the app-maven-plugin repo).

In practice, this means that this integration test reuses app-maven-plugin's git repo and will only work if the plugin code is managed by some kind of repo supported by gen-repo-info-files. Let me know if you have any suggestions for a better approach here.

src/test/java/com/google/cloud/tools/maven/GenRepoInfoFileMojoTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Test
+  public void testExecute() throws MojoFailureException, MojoExecutionException {
+    genMojo.ignoreErrors = true;
+    genMojo.execute();
+
+    verify(genMock).generate(genMojo);
+  }
+
+  @Test(expected = MojoExecutionException.class)
```
joaoandremartins: I chose not to inspect the exception's text to use this simple (and recommended I think?) notation. If you think something else could be throwing MojoExecutionExceptions, I can add a check for text.

joaoandremartins:
OK, I think this should be ready for final review.

I tested both new integration tests and they both succeeded locally.

src/test/java/com/google/cloud/tools/maven/GenRepoInfoFileMojoTest.java
meltsufin: This is fine for now.

src/test/java/com/google/cloud/tools/maven/it/GenRepoInfoFileMojoIntegrationTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Ensures that the genRepoInfoFile goal is automatically called for a flexible project
+   * deployment.
+   */
+  @Test
+  public void testGenerateFlex() throws VerificationException, IOException, ProcessRunnerException {
+    Verifier verifier = new FlexibleVerifier("testGenRepoInfoFile_flex");
+
+    verifier.executeGoal("appengine:deploy");
```
meltsufin: The goal is bound to `prepare-package`. So, I think you should test that it happens then. In addition, deploy is too expensive. We don't want to do it in our integration tests, if we don't have to.
```Java
verifier.executeGoal("package");
```

src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.maven.plugins.annotations.Execute;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+
+import java.io.File;
+import java.nio.file.Paths;
+
+/**
+ * Generates source context files.
```
joaoandremartins: Generates a source-context.json and source-contexts.json, so I guess it's plural. Though one of these is provided for legacy support, I think. gcloud stage only creates one file, source-context.json IIRC.

src/test/java/com/google/cloud/tools/maven/it/GenRepoInfoFileMojoIntegrationTest.java
joaoandremartins: Added a test to check if goal is ran with package.

I would actually prefer to keep the deploy test, as source contexts on flex deployments were a big motivator for doing this. In this instance, I want to ensure that for a project with the executions bit in its pom.xml, the goal is automatically ran and the files are correctly added.

joaoandremartins:
@meltsufin PTAL

src/test/java/com/google/cloud/tools/maven/it/GenRepoInfoFileMojoIntegrationTest.java
meltsufin: Do you want to just add the assertion in `DeployMojoIntegrationTest.testDeployFlexible()`?

joaoandremartins: Good idea. Done.

src/main/java/com/google/cloud/tools/maven/GenRepoInfoFileMojo.java
```diff
@@ -0,0 +1,85 @@
...
+  protected boolean ignoreErrors;
+
+  @Override
+  public void execute() throws MojoExecutionException, MojoFailureException {
+    try {
+      getAppEngineFactory().genRepoInfoFile().generate(this);
+    } catch (AppEngineException aee) {
+      if (!ignoreErrors) {
+        throw new MojoExecutionException("An error occurred while generating source context files."
+            + " To ignore source context generation errors and proceed to deployment, use the "
```
meltsufin: Sorry, it should be set to 'true', not false.

joaoandremartins: In the unlikely event of there being an error, I think it would be better to let the user know about it more visibly, rather than moving on and possibly hiding the exception in the middle of all the other logging rubble, no?

joaoandremartins: @meltsufin PTAL

joaoandremartins:
@patflynn @loosebazooka do you guys want to take a look here? no worries if no time, we can always add more stuff later

joaoandremartins:
Brad Daniels from CDB approved this offline, so I think we're ready to merge!

PR #132, opened 2017-02-03
Allow app.yaml to be generated by gcloud when deploying

### LINKED ISSUE #131: Allow app.yaml to be automatically generated when deploying

meltsufin:
Fixes #131.

This is pending fix in appengine-plugins-core: https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/317

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/DeployMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/DeployMojo.java
@@ -96,6 +96,8 @@ public class DeployMojo extends StageMojo implements DeployConfiguration {
     File appYamlFile = new File(getStagingDirectory() + "/app.yaml");
     if (deployables.size() == 0 && appYamlFile.exists()) {
       deployables.add(appYamlFile);
+    } else {
+      deployables.add(stagingDirectory);
     }
 
     getAppEngineFactory().deployment().deploy(this);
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/maven/DeployMojo.java
```diff
@@ -93,9 +93,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
     super.execute();
 
-    if (deployables.size() == 0) {
-      deployables.add(
-          new File(getStagingDirectory() + "/app.yaml"));
+    File appYamlFile = new File(getStagingDirectory() + "/app.yaml");
+    if (deployables.size() == 0 && appYamlFile.exists()) {
+      deployables.add(appYamlFile);
+    } else {
+      deployables.add(stagingDirectory);
```
joaoandremartins: Why don't we stage an app.yaml file instead, if none exists?

meltsufin: I don't think we want to maintain yet another app.yaml generation template.
Here we're just letting the Cloud SDK generate the app.yaml based on its fingerprinting logic.

joaoandremartins: I don't feel very strongly either way, so sure

chanseokoh: Question: you use `getStagingDirectory()` above, but here, it's just `stagingDirectory`. This makes me wonder if there is a difference.

meltsufin: No difference, but good catch, the code should use consistent access.

PR #141, opened 2017-02-27
Update snapshot version after release

loosebazooka:


pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-maven-plugin</artifactId>
-  <version>1.1.1-beta-SNAPSHOT</version>
+  <version>1.3.0-SNAPSHOT</version>
```
meltsufin: Maybe `1.2.1-SNAPSHOT`?

loosebazooka: yeah, that is better

PR #149, opened 2017-03-08
Adds a parent pom

joaoandremartins:
Allows genRepoInfoFile to be ran without an executions block on the client pom, but can be used for a multitude of other things, such as default configs for certain goals, etc.

This change still requires documentation changes.


pom.xml
```diff
@@ -5,7 +5,10 @@
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-maven-plugin</artifactId>
   <version>1.2.2-SNAPSHOT</version>
-  <packaging>maven-plugin</packaging>
+  <modules>
+    <module>app-maven-plugin-parent</module>
+  </modules>
+  <packaging>pom</packaging>
```
joaoandremartins: I'll actually need to create a separate module to be the maven-plugin, and the top-level pom.xml will be the "aggregator" module.

app-maven-plugin-parent/pom.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.google.cloud.tools</groupId>
+    <artifactId>app-maven-plugin-parent</artifactId>
```
elharo: Why appmaven-plugin here and appengine-maven-plugin below?

pom.xml
meltsufin: So, is this PR still in progress then?

loosebazooka:
This could be a nice convenience, but I think @meltsufin made a good point about parent pom usage. I'm not convinced this is the way we should go about doing this vs documentation.

Just curious how would the release process work for this?

joaoandremartins:
I just made a few changes, so the proposal now is to have an aggregator pom "appengine-maven", which contains modules "appengine-maven-parent" and "appengine-maven-plugin". I can't really make up my mind on what's best, whether to move everything around to include the parent pom module in this repo, create a new repo just for the parent pom, not add the parent pom, etc.

To your concerns, using the parent pom is strictly optional. It comes in handy if we want to add template configuration to an app. It's [what Spring Boot uses](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-starters/spring-boot-starter-parent/pom.xml) to calls its build process hook without explicitly having the executions block in the pom.xml.

I'm not sure what you mean with the release process. I guess we just release it the same way as before, just using the aggregator pom.xml?

pom.xml
```diff
@@ -3,9 +3,13 @@
...
   <groupId>com.google.cloud.tools</groupId>
-  <artifactId>appengine-maven-plugin</artifactId>
+  <artifactId>appengine-maven</artifactId>
   <version>1.2.2-SNAPSHOT</version>
-  <packaging>maven-plugin</packaging>
+  <modules>
+    <module>appengine-maven-parent</module>
+    <module>appengine-maven-plugin</module>
+  </modules>
+  <packaging>pom</packaging>
```
meltsufin: Since this is a pom now, I would move the whole `<dependencies>` block in this parent to `appngine-maven-plugin\pom.xml`.

appengine-maven-parent/pom.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.google.cloud.tools</groupId>
+    <artifactId>appengine-maven-parent</artifactId>
+    <version>1.2.2-SNAPSHOT</version>
```
meltsufin: Specify `<name>` as well.

appengine-maven-plugin/pom.xml
```diff
@@ -0,0 +1,222 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>appengine-maven</artifactId>
+        <groupId>com.google.cloud.tools</groupId>
+        <version>1.2.2-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>appengine-maven-plugin</artifactId>
+    <packaging>maven-plugin</packaging>
```
meltsufin: Specify `<name>` as well.

chanseokoh:
> I'm not sure how this will work. Will the client application need to directly inherit from our parent pom and get the default plugin configuration without even having to explicitly specify the plugin? If that's the idea, a major issue with it is that most people will want to inherit from their own parent pom (if any), and you can't inherit from multiple parent poms.

I'm no expert in Maven, but even if every plugin defines this kind of "parent" pom Joao proposes, I think people can still chain these poms in sequence even if they have their own parent pom. That said, I see this kind of "parent" pom as a super pom for a certain plugin, much like Maven by itself has the super POM that every pom inherits from; all child poms in the hierarchy will be affected, and it's not possible that some subset of children can selectively inherit it.

In that sense, I think this "super" pom for our plugin should be to provide absolutely global defaults, not just for a convenience that provides a quick means to enable certain features. I haven't looked into how Spring Boot set up this model, but conceptually, it does not make sense to me if these "super" poms are not for global defaults.

meltsufin:
You can override the plugin configuration in the child pom.
In any case, many apps already have a parent pom, and somewhere up the chain of their ancestor poms there is pom they do not control. So, in that situation this parent cannot be used.

joaoandremartins:
Just to clarify, the reason I'm proposing this change is mainly to not have to specify the <executions> block. Then again, that block would need to be replaced with another one, <parent>, so I don't feel very strongly about this. Would be good to hear from @patflynn , he showed interest in this earlier.

@meltsufin I also thought this would go well on our new project templates, but not too sure what else we could gain from it immediately (besides not having to specify the <executions> block to generate source context)

meltsufin:
@saturnism

patflynn:
I thought that the maven-git-commit-id-plugin was doing something like this without a parent pom, but I have verified it.

Take a look here: https://github.com/ktoso/maven-git-commit-id-plugin/blob/8cc014bba80223fa1ed9636abbb252cdb836c399/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L70

It appears to use an annotation to bind the mojo to a lifecycle phase by default.

I am a Maven mojo ignoramus so I may be reading it wrong.

joaoandremartins:
FYI I tried to bind the goal to a default phase and couldn't get it to automatically run without explicitly calling the goal.

patflynn:
aha. It looks like you still need an execution for it to run. Looking at the springboot parent pom it actually configures the git-commit-id-plugin to execute. 

I'm not quite sold on this change though.  For flex users on springboot they'd neither want to use our parent pom (since they want to use the springboot one), or use our source context files since the git.properties file will work. 

For standard it could turn out to be interesting, but standard src context works already work.. and we want to be careful not to confuse things since this plugin supports both sets of users (for now).

joaoandremartins:
Alright, I don't feel very strongly about this approach, either. I agree the Maven pom hierarchy has its limitations.

patflynn:
It's an interesting idea that's worth keeping in mind for standard users.  You could for instance pre-configure the build to perform staging, and deploy to deploy to app engine.

@loosebazooka if we ever split out the plugins this could be worth looking at.

PR #156, opened 2017-03-24
Allow licenses use 2017 as well as 2016 for copyright year

akerekes:


chanseokoh:
I'm just wondering if `.` (dot) should also be escaped.

akerekes:
@chanseokoh yeah, it should be.

patflynn:
Shouldn't this just be 2017?

meltsufin:
For new code, probably, but I'm not sure about existing files. Do you just go every Jan 1st, and change all headers everywhere?

akerekes:
http://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

>The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made.

patflynn:
No we don't change all the headers. We just update our copyright template
to 2017.
https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/.idea/copyright/Google_Copyright.xml

IntelliJ has the habit of updating the copyright though on files that you
change in a commit.

PR #157, opened 2017-03-24
Fixes #155 - Add deploy mojos for config files

### LINKED ISSUE #155: Add deploy mojos for config files
```
appengine:deployCron
appengine:deployDispatch
appengine:deployDos
appengine:deployIndex
appengine:deployQueue
```

configuration through :
```
<appEngineDirectory/> // for configuration file locations
<project/> // to specify gcloud project
```

akerekes:


loosebazooka:
@patflynn proposed some naming changes here which might affect this CL, nothing major though.

src/main/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojo.java
```diff
@@ -0,0 +1,68 @@
...
+  public AbstractSingleYamlDeployMojo() {
+    super();
+  }
+
+  @Override
+  public void execute() throws MojoExecutionException, MojoFailureException {
+    // execute stage
+    super.execute();
+  
+    if (isStandardStaging()) {
```
loosebazooka: So I think we're changing this parameter (from appEngineDirectory), but anyway :

I wonder if the correct way of doing optional defaults (flex vs standard) is to unset the default for appEngineDirectory in "StagingMojo" and only set the default here if appEngineDirectory == null to either the standard or flex location.

src/test/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojoTest.java
```diff
@@ -89,7 +90,8 @@ public void testDeployStandard()
     mojo.execute();
 
     // verify
-    assertEquals(tempFolder.getRoot().getAbsolutePath() + "/staging/WEB-INF/appengine-generated",
+    assertEquals(Paths.get(tempFolder.getRoot().getAbsolutePath(),
```
loosebazooka: can you use paths.isSameFile ?

src/main/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojo.java
loosebazooka: ignore this.

akerekes:
@loosebazooka PTAL

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -223,6 +223,15 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     }
   }
 
+  private void initializeAppEngineWebXml() throws MojoExecutionException {
```
loosebazooka: Does this give us anything? It's just one line?

src/test/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojoTest.java
```diff
@@ -0,0 +1,116 @@
...
+    // create appengine-web.xml to mark it as standard environment
+    File appengineWebXml = new File(tempFolder.newFolder("source", "WEB-INF"), "appengine-web.xml");
+    appengineWebXml.createNewFile();
+    Files.write("<appengine-web-app></appengine-web-app>", appengineWebXml, Charsets.UTF_8);
+
+    // invoke
+    mojo.execute();
+
+    // verify
+    assertTrue(java.nio.file.Files.isSameFile(Paths.get(tempFolder.getRoot().getAbsolutePath(),
```
loosebazooka: oh you know what, I might be wrong here, the actually error description provided by comparing the strings might be much better

src/test/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojoTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+
+@RunWith(MockitoJUnitRunner.class)
+public abstract class AbstractSingleYamlDeployMojoTest<M extends AbstractSingleYamlDeployMojo> {
```
loosebazooka: Obviously this is personal preference, but can we write this as a test fixture instead of a superclass? I find it easier to trace through and debug tests that way, and I'm not sure how I feel about `@Test` inheritance from a readability standpoint.

I'm trying to figure out how to exactly do this, but perhaps DeployCronMojo would get the common code by instantiating SingleYamlDeployMojoTestFixture<M>(tmpDir) which would mock all required things (and probably have to call initMocks on itself). I honestly have no idea. Do you think this makes sense though?

src/test/java/com/google/cloud/tools/maven/util/SingleYamlDeployTestHelper.java
```diff
@@ -0,0 +1,97 @@
...
+  private CloudSdkAppEngineFactory factoryMock;
+
+  @InjectMocks
+  protected M mojo;
+
+  private boolean isStandard;
+
+  private TemporaryFolder temporaryFolder;
+
+  public SingleYamlDeployTestHelper(M mojo, TemporaryFolder temporaryFolder, boolean b) {
```
loosebazooka: I see the instantiators below, should we make this private?

loosebazooka: but maybe also rename `b` to `isStandard` or something?

src/test/java/com/google/cloud/tools/maven/util/SingleYamlDeployTestHelper.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.rules.ExternalResource;
+import org.junit.rules.TemporaryFolder;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import java.io.File;
+import java.io.IOException;
+
+public class SingleYamlDeployTestHelper<M extends AbstractSingleYamlDeployMojo>
```
loosebazooka: External Resource thing is cool, would it make sense to have this as separate flex/standard fixtures?, it does seem like most of the heavy lifting is in the branches in the before and after?

akerekes:
@loosebazooka PTAL

src/test/java/com/google/cloud/tools/maven/util/SingleYamlStandradDeployTestHelper.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public void before() throws IOException {
+    mojo.setStagingDirectory(temporaryFolder.newFolder("staging"));
+    mojo.setSourceDirectory(temporaryFolder.newFolder("source"));
+    MockitoAnnotations.initMocks(this);
+
+    // create appengine-web.xml to mark it as standard environment
+    File webInfDirectory = mojo.getSourceDirectory().toPath().resolve("WEB-INF").toFile();
+    Assert.assertTrue(webInfDirectory.mkdir());
```
loosebazooka: I don't think we actually need to assert these  as junit asserts? do we?

src/main/java/com/google/cloud/tools/maven/DeployCronMojo.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apache.maven.plugins.annotations.Execute;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+
+/**
+ * Stage and deploy cron.yaml to Google App Engine standard or flexible environment.
+ */
+@Mojo(name = "deployCron")
+@Execute(phase = LifecyclePhase.PACKAGE)
```
patflynn: Does this bind executions to the package phase by default?

patflynn:
#155

PR #160, opened 2017-03-31
Update code to use latest plugins-core version

### LINKED ISSUE #158: dev appserver 1 stuff
Include new devappserver1 stuff from plugins-core.

Add `<devAppServerVersion>` configuration to pick between 1 and 2. It only accepts 1 and 2.
I think you might need to add a few params here and there and maybe change "storagePath" to file, you'll see it all.

Might want to talk with @elharo @patflynn on naming so we're somewhat consistent across platforms. (naming for choosing 1 or 2)

akerekes:
This is WIP, but I put it out there for early feedback.
Needs tests and a new release of the `appengine-plugins-core`.
Builds are failing because https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/368 hasn't been merged yet.


src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -40,11 +43,18 @@
   /**
    * Path to a yaml file, or a directory containing yaml files, or a directory containing
    * WEB-INF/web.xml.
+   *
+   * @deprecated 1.2.2
```
patflynn: it'd be nice to explain here what the user needs to do to replace their app.yaml configs here.

src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -80,7 +90,7 @@
    * Path to the data (datastore, blobstore, etc.) associated with the application. (default: None)
    */
   @Parameter(alias = "devserver.storagePath", property = "app.devserver.storagePath")
-  protected String storagePath;
+  protected File storagePath;
```
patflynn: @loosebazooka

Does maven automatically convert the string to file? Will this have any effect on the end user configs?

src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -201,11 +211,56 @@
...
+            + " <appYamls> is deprecated, use <services> only.");
+      }
+    }
+  }
+
+  private boolean servicesUsesTheDefaultValue() throws MojoExecutionException {
+    MavenProject project = (MavenProject)getPluginContext().get("project");
+    Plugin ourPlugin = project.getPlugin("com.google.cloud.tools:appengine-maven-plugin");
+    Object configuration = ourPlugin.getConfiguration();
+    if (configuration instanceof org.apache.maven.shared.utils.xml.Xpp3Dom) {
```
patflynn: this seems very convoluted just to figure out if this setting was provided by the user.

src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -201,11 +211,56 @@
...
+        getLog().warn("The value of <appYamls> will override the default value of <services>");
+        services = appYamls;
+      } else {
+        throw new MojoExecutionException("Both <appYamls> and <services> are defined."
+            + " <appYamls> is deprecated, use <services> only.");
+      }
+    }
+  }
+
+  private boolean servicesUsesTheDefaultValue() throws MojoExecutionException {
```
joaoandremartins: Some Javadoc explaining what's going on here could be helpful.

joaoandremartins:
Is it worth adding a todo or something to delete the appYamls bit eventually?
Do we want to do that now?

src/main/java/com/google/cloud/tools/maven/RunMojo.java
akerekes: Yeah, although it'd be nice to have a better reference of what is expected here by Maven. According to the [docs](http://maven.apache.org/ref/3.2.1/maven-plugin-api/plugin.html):
>Specify the version when the Mojo was deprecated to the API. Similar to Javadoc deprecated. This will trigger a warning when a user tries to configure a parameter marked as deprecated.

However this doesn't seem to work in my local development, so we could just document the reason along with the version, and display the warning in the `execute()` method as it is now.

src/main/java/com/google/cloud/tools/maven/RunMojo.java
akerekes: It does, and I'm not aware of any effect. They provide a path string in both cases. https://maven.apache.org/guides/plugin/guide-java-plugin-development.html#Files_and_Directories

src/main/java/com/google/cloud/tools/maven/RunMojo.java
akerekes: I've simplified it a bit after a search and seeing that all current maven code (in my Eclipse workspace) uses only the class form Plexus. However this may break in the future if the plan to rewrite the plexus utils finishes and Maven plugins migrate.

src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -36,15 +39,22 @@
...
-  // TODO: the name doesn't fit when you pass a build dir, maybe yamlOrWarPath?
+  // TODO remove this: https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/162
   /**
    * Path to a yaml file, or a directory containing yaml files, or a directory containing
    * WEB-INF/web.xml.
+   *
+   * @deprecated 1.2.2
    */
   @Parameter(alias = "devserver.appYamls", property = "app.devserver.appYamls",
+      defaultValue = "${project.build.directory}/${project.build.finalName}")
```
akerekes: Yes, we should, I'll fix it

loosebazooka:
I think we want to allow an option of switching back and forth between the v1 and v2 implementations in build tools.

For gradle I'm using a serverVersion parameter which can be `1` or `2-beta`
so a user (in gradle) would use it like this :
```
appengine {
  run {
    serverVersion = "1"
  }
}

and it fails if they use anything else.

akerekes:
I'm going to repurpose this PR from adding devappserver1 support to "migrate plugin to new plugins-core version" and add devappserver1 in a separate one to keep them smaller.

patflynn:
What do you guys think about 'v1' and 'v2-alpha'?   I don't think the v2
server is really feature complete at this stage for Java purposes so 'beta'
feels premature.

loosebazooka:
Yeah that's fine

akerekes:
I'd leave out the `v` prefix, but otherwise `1` and `2-alpha` sounds good.

Issue #164, opened 2017-04-05
Stop should generate a configuration based on dev appserver 1 or 2-alpha.


loosebazooka:
dev appserver1 uses the host and port to stop the server
dev appserver2 uses the adminHost and adminPort to stop the server

Currently I believe the StopMojo only has access to the adminHost and adminPort, but it needs access to the host and port for devAppServer1.

loosebazooka:
closed by #165 ?

PR #165, opened 2017-04-06
Fixes: #158 - dev appserver 1 stuff and #164 - Stop should generate a configuration based on dev appserver 1 or 2-alpha.

akerekes:


akerekes:
@patflynn @loosebazooka PTAL

pom.xml
```diff
@@ -128,6 +128,13 @@
       <scope>test</scope>
     </dependency>
 
+	<dependency>
+	  <groupId>pl.pragmatists</groupId>
+	  <artifactId>JUnitParams</artifactId>
+	  <version>1.0.6</version>
+	  <scope>test</scope>
+	</dependency>
```
loosebazooka: spacing?

src/main/java/com/google/cloud/tools/maven/AppEngineFactory.java
```diff
@@ -27,17 +27,38 @@
  */
 public interface AppEngineFactory {
 
+  /**
+   * Supported dev app server versions.
+   */
+  public enum SupportedVersion {
```
loosebazooka: since this isn't scoped to the server explicitly?, should it SupporterServerVersion?

src/test/java/com/google/cloud/tools/maven/RunAsyncMojoTest.java
```diff
@@ -40,13 +46,22 @@
...
   private RunAsyncMojo runAsyncMojo;
 
+  @Before
+  public void setUp(){
+    MockitoAnnotations.initMocks(this);
+  }
+
   @Test
-  public void testRunAsync() throws MojoFailureException, MojoExecutionException, IOException {
+  @Parameters({"1,V1", "2-alpha,V2ALPHA"})
```
loosebazooka: That's pretty cool

src/test/java/com/google/cloud/tools/maven/RunAsyncMojoTest.java
```diff
@@ -40,13 +46,22 @@
...
+  public void testRunAsync(String version, SupportedVersion mockVersion)
+      throws MojoFailureException, MojoExecutionException, IOException {
     final int START_SUCCESS_TIMEOUT = 25;
 
     // wire up
+    runAsyncMojo.devserverVersion = version;
```
loosebazooka: This part is a little confusing. why do we need both version and mockVersion?

akerekes: The `mockVersion` is used to configure `factoryMock` to respond only to the appropriate `devServerRunAsync()` method call. `version` is used by the class under test to call the mocked `devServerRunAsync()`.

src/main/java/com/google/cloud/tools/maven/AppEngineFactory.java
akerekes: done

pom.xml
akerekes: done

src/test/java/com/google/cloud/tools/maven/RunAsyncMojoTest.java
loosebazooka: But the two values are the same, no? Shouldn't we be able to derive one from the other? Or is that bad for this test?

PR #172, opened 2017-04-12
update the version after release

akerekes:


pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-maven-plugin</artifactId>
-  <version>1.2.2-SNAPSHOT</version>
+  <version>1.3.0-SNAPSHOT</version>
```
meltsufin: not 1.2.3?

akerekes: we've release 1.3.0-rc1

loosebazooka: There's some new features and stuff.

loosebazooka:
@akerekes Update the change log too in this PR if you can.

USER_GUIDE.md
```diff
@@ -0,0 +1,86 @@
...
+You can now deploy the cron/doc/etc. configuration files separately using the new goals:
+
+* `deployCron`
+* `deployDispatch`
+* `deployDos`
+* `deployIndex`
+* `deployQueue`
+
+## Source directory
+_For GAE Flexible projects_ The deployment source directory can be overridden by setting the `appEngineDirectory` parameter in the deploy configuration.
```
loosebazooka: Does _For_ need to be italicizied?

akerekes: I want to somehow emphasize, that it's only for flex projects. Maybe a different formatting/wording would be better? I meant it as a separate label, not part of the sentence starting with "The deployment..."

loosebazooka: Oh right, never mind that makes sense. I was thinking "For _GAE Flexible projects_", but I think what you have makes more sense.

PR #174, opened 2017-04-13
Fixes #173 - Setting appEngineDirectory for GAE Standard projects has no effect

### LINKED ISSUE #173: Setting appEngineDirectory for GAE Standard projects has no effect
Currently we override the `appEngineDirectory` parameter with `<staging-dir>/WEB-INF/appengine-generated` so even if `appEngineDirectory` is set in the pom.xml, it will have no effect when users run deployCron/Dos/etc. on a GAE Standard project.

We should either allow them to override this, or emit a warning message about this if they try to set it.

akerekes:
The initial design to check whether a parameter was incorrect. It did not take into account whether the parameter was set using its alias or via a system property. Moreover the logic for parameter default value in Maven API does not support our need of default value based on condition, i.e. use A is it's a GAE Standard project, and use B if it's a GAE Flexible project.
So I moved the default value handling into the `execute()` method.

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojo.java
@@ -43,13 +43,25 @@ public abstract class AbstractSingleYamlDeployMojo
   public void execute() throws MojoExecutionException, MojoFailureException {
     // execute stage
     super.execute();
+    doDeploy(getAppEngineFactory().deployment(), this);
+  }
   
+  /**
+   * Sets {@code appEngineDirectory} based on whether the project is GAE Standard or Flexible if
+   * the user has not set a value explicitly.
+   *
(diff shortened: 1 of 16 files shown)


src/main/java/com/google/cloud/tools/maven/AbstractSingleYamlDeployMojo.java
```diff
@@ -43,13 +43,25 @@ public AbstractSingleYamlDeployMojo() {
...
     // execute stage
     super.execute();
+    doDeploy(getAppEngineFactory().deployment(), this);
+  }
   
+  /**
+   * Sets {@code appEngineDirectory} based on whether the project is GAE Standard or Flexible if
+   * the user has not set a value explicitly.
+   *
+   * <p>For Standard it uses {@code <stagingDirectory>/WEB-INF/appengine-generated}, for Flexible
```
loosebazooka: why html and javadocs styles? `@code` vs `<code>`

akerekes: To avoid escaping for the sake of readability of code.

PR #175, opened 2017-04-14
more integration tests for RunMojo

akerekes:


src/test/java/com/google/cloud/tools/maven/it/RunMojoIntegrationTest.java
```diff
@@ -86,6 +97,34 @@ public void run() {
     assertEquals("Hello from the App Engine Standard project.", urlContent.toString());
     verifier.verifyErrorFreeLog();
     verifier.verifyTextInLog("Dev App Server is now running");
+    verifier.verifyTextInLog("Module instance " + expectedModuleName + " is running");
+  }
+
+  @SuppressWarnings("unused") // used for testRun()
```
loosebazooka: I don't understand why it would think it was unused, if it was in fact used? Is there some annotation somehwere that pulls it in?

src/test/java/com/google/cloud/tools/maven/it/RunMojoIntegrationTest.java
```diff
@@ -86,6 +97,34 @@ public void run() {
...
+  private Object[] parametersForTestRun() {
+    List<Object[]> result = new ArrayList<>();
+    for (SupportedDevServerVersion serverVersion : SupportedDevServerVersion.values()) {
+      for (String[] profileAndServiceName : new String[][]{
+        { "", "standard-project"},
+        { "base-it-profile,appyamls", "standard-project-appyamls"},
+        { "base-it-profile,services", "standard-project-services"}
+        } ) {
+        result.add(
```
loosebazooka: I feel like I would prefer this code to just build the Object[] directly instead of through indirection via the String[][]
```
result.add(newObject[] {serverVersion, new String[]{"",""}, "standard-project"})
result.add(newObject[] {serverVersion, new String[]{"base-it-profile","appyamls"}, "standard-project-appyamls"})
result.add(newObject[] {serverVersion, new String[]{"base-it-profile","services"}, "standard-project-services"})
```
or maybe through a helper
```
paramHelperOrSomething(serverVersion, profile, subprofile, service)
```

whaddya think?

src/test/java/com/google/cloud/tools/maven/it/RunMojoIntegrationTest.java
akerekes: JUnitParameters uses the convention that if the test method is called fooBar, then the method providing the parameters is called parametersForFooBar (a convention that can be overridden, but that wouldn't change the fact that Eclipse marks it as unused).

src/test/java/com/google/cloud/tools/maven/it/RunMojoIntegrationTest.java
akerekes: I agree, it's pretty hard to read what's going on. I think I prefer the first option for now.

src/test/java/com/google/cloud/tools/maven/it/RunMojoIntegrationTest.java
loosebazooka: Ah okay, can you add a comment just linking it to the test method? like`@{link whatever}`?

PR #176, opened 2017-04-19
update user guide for 1.3.0

akerekes:


src/main/java/com/google/cloud/tools/maven/DeployMojo.java
```diff
@@ -52,6 +52,8 @@
 
   /**
    * Deploy with a specific Docker image. Docker url must be from one of the valid gcr hostnames.
+   * 
+   * <p><i>Supported only for devserver version 2-alpha.</i>
```
loosebazooka: I think this comment mean to be "only for flex?"

PR #182, opened 2017-04-24
Run plugin

akerekes:
In preparation for separate run/multimodule run/std deploy/flex deploy plugins, I've separated the run and deploy functions into 2 plugins, and extracted some common code into its own module.


loosebazooka:
I guess we don't need this anymore.

PR #183, opened 2017-05-01
update plugins-core to 0.3.2, add <environment> parameter

akerekes:
I don't particularly like how the <environment> parameter is tested in the integration test, but until we add more thorough testing of parameters to integration tests, it'll be OK.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -214,7 +214,7 @@ To switch back to the Dev App Server v2-alpha (that was default in version < 1.3
 
 ### How can I pass environment variables to the Dev Appserver (both v1 and v2-alpha)?
 
-You can debug the Dev App Server v1 using the jvmFlags:
+You can pass environment variables directly to the Dev App Server:
 
 ```XML
 <configuration>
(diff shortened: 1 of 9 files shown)


loosebazooka:
presumably, just ensuring the parameter is configured and set is enough no?

USER_GUIDE.md
```diff
@@ -84,6 +84,7 @@ Valid for versions "1" and "2-alpha":
 | `startSuccessTimeout` | Amount of time in seconds to wait for the Dev App Server to start in the background. |
 | `devserverVersion`    | Server versions to use, options are "1" or "2-alpha" |
 | `services`            | List of services to run |
+| `environment`        | Environment variables to pass to the Dev App Server process |
```
loosebazooka: I think this should be sorted alphabetical?

USER_GUIDE.md
```diff
@@ -211,6 +212,18 @@ To switch back to the Dev App Server v2-alpha (that was default in version < 1.3
 </configuration>
 ```
 
+### How can I pass environment variables to the Dev Appserver (both v1 and v2-alpha)?
+
+You can debug the Dev App Server v1 using the jvmFlags:
```
loosebazooka: Update this comment?

You can now pass environment variables directly to the Dev App Server.

Issue #217, opened 2017-10-20
Upgrade appengine-plugins-core and make a new release


chanseokoh:
Recent appengine-plugins-core versions fix some problems.

chanseokoh:
1.9.59 released. Closing. BTW, 1.9.59 should fix #229 (when running a single service).

coollog:
Hey chanseok, we haven't yet released a new version yet. The latest version is still 1.3.1. We will release 1.3.2 with the upgraded appengine-plugins-core soon.

chanseokoh:
Oops. I took it wrong.

PR #219, opened 2017-10-24
Upgrades appengine-plugins-core to 0.3.9 and adds support for new add…

coollog:
First step of #217 . This adds support for additional arguments to pass to dev app server that was added in https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/432.


loosebazooka:
we need up to update the cloud docs too

coollog:
kokoro-macos build should be fixed once #221 is merged.

PR #221, opened 2017-10-24
Correctly updates gcloud in continuous_mac.sh

### LINKED ISSUE #222: continuous_mac.sh uses old version of gcloud
`continuous_mac.sh` downloads the latest Cloud SDK but does not actually set the path of gcloud. This can be seen by failures on kokoro-macos such as:

```
testRunAsyncV2Alpha: [ERROR] Failed to execute goal com.google.cloud.tools:appengine-maven-plugin:1.3.2-SNAPSHOT:start (default-cli) on project standard-project: Execution default-cli of goal com.google.cloud.tools:appengine-maven-plugin:1.3.2-SNAPSHOT:start failed: Requires version 171.0.0 or later but found version 159.0.0 -> [Help 1]
```

coollog:
Fixes #222

### DIFF
--- a/kokoro/continuous_mac.sh
+++ b/kokoro/continuous_mac.sh
@@ -10,9 +10,7 @@ GOOGLE_CLOUD_SDK_HOME=/Users/kbuilder/google-cloud-sdk
 "$GOOGLE_CLOUD_SDK_HOME"/bin/gcloud components install app-engine-java
 
 # Sets gcloud to use the downloaded SDK.
-GCLOUD_PATH=$(which gcloud)
-rm $GCLOUD_PATH
-ln -s "$GOOGLE_CLOUD_SDK_HOME"/bin/gcloud $GCLOUD_PATH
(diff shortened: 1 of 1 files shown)


loosebazooka:
can you explain?

coollog:
Oh this is a test PR, not for review haha. I'm trying to fix the breaking macos build in https://github.com/GoogleCloudPlatform/app-maven-plugin/pull/219

coollog:
Okay, now this is a real PR, as this is what we want to do. The continuous_mac.sh script downloaded the Cloud SDK but did not actually have it be used. This fixes that by having the correct symlink to the downloaded gcloud.

loosebazooka:
should we just use the cloudSDK on the mac already, similar to the linux build?

coollog:
The Cloud SDK on the mac is outdated: #222

chanseokoh:
If so, perhaps we could just do `gcloud update components` with the built-in Cloud SDK.

coollog:
Okay just doing that worked! Thanks!

PR #230, opened 2017-11-07
Release on v1.3.2

coollog:
Release on `v1.3.2`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-maven-plugin</artifactId>
-  <version>1.3.2-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


chanseokoh:
Compared to https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/182, why did that one not have SNAPSHOT?

coollog:
That is definitely an error. The commits made there are done by the gradle-release plugin. I am not sure why it did not put -SNAPHOT for the new version. I checked running the gradle-release plugin and it works fine (adds -SNAPSHOT) so it might have been a mismatch between the release branch version and the gradle.properties version that caused the gradle-release plugin to not update the version correctly. I filed https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/187/files to fix the version.

PR #253, opened 2018-03-14
Add skeleton for cloud sdk download/check

### LINKED ISSUE #256: Use path of downloaded cloud sdk in sdk factory when cloudSdkPath isn't configured
Currently: 
```
return cloudSdkFactory
        .cloudSdkBuilder()
        .sdkPath(mojo.getCloudSdkPath())
        ...
```
When the cloud sdk is downloaded, the new installed directory should replace the null configured directory.

TadCordle:
Fixes #256

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
@@ -30,12 +30,9 @@ import com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineStandardStagin
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdkGenRepoInfoFile;
 import com.google.cloud.tools.appengine.cloudsdk.process.NonZeroExceptionExitListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
-
 import org.apache.maven.plugin.logging.Log;
 
-/**
- * Factory for App Engine dependencies.
- */
+/** Factory for App Engine dependencies. */
 public class CloudSdkAppEngineFactory implements AppEngineFactory {
 
   protected CloudSdkFactory cloudSdkFactory;
@@ -74,8 +71,8 @@ public class CloudSdkAppEngineFactory implements AppEngineFactory {
     return createDevServerForVersion(version, defaultCloudSdkBuilder().build());
   }
 
-  private AppEngineDevServer createDevServerForVersion(SupportedDevServerVersion version,
-      CloudSdk cloudSdk) {
+  private AppEngineDevServer createDevServerForVersion(
+      SupportedDevServerVersion version, CloudSdk cloudSdk) {
     switch (version) {
       case V1:
         return cloudSdkFactory.devServer1(cloudSdk);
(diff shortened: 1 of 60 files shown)


TadCordle:
Gonna refactor a bit.

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -29,8 +29,12 @@
 public abstract class CloudSdkMojo extends AbstractMojo {
 
   /** Optional parameter to configure the location of the Google Cloud SDK. */
-  @Parameter(property = "cloudSdkPath", required = false)
-  private File cloudSdkPath;
+  @Parameter(property = "cloudSdkHome", required = false)
```
loosebazooka: I wouldn't remove cloudSdkPath completely, maybe use the strategy used for appYamls in RunMojo that kind of shows a deprecation notice? It might not be best if we just swapped this name on users without warning.

TadCordle: Gonna change it back for now.

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -48,10 +52,14 @@ public String getArtifactVersion() {
     return pluginDescriptor.getVersion();
   }
 
-  public Path getCloudSdkPath() {
+  public Path getcloudSdkPath() {
```
loosebazooka: I think you need to capitalize the `C` in `cloud` here.

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
```diff
@@ -88,6 +96,15 @@ public String getCompileTargetVersion() {
     return javaVersion;
   }
 
+  public void downloadCloudSdk() {
```
loosebazooka: Would it make sense to put these in defaultCloudSdkBuilder?

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -102,12 +102,17 @@ public GenRepoInfoFile genRepoInfoFile() {
...
+    } else if (mojo.getCloudSdkVersion() != null) {
+      mojo.checkCloudSdk();
+    }
 
     ProcessOutputLineListener lineListener = new DefaultProcessOutputLineListener(mojo.getLog());
 
     return cloudSdkFactory
         .cloudSdkBuilder()
-        .sdkPath(mojo.getCloudSdkPath())
+        .sdkPath(mojo.getcloudSdkPath())
```
loosebazooka: Can't find where this uses the downloaded SDK's home path?

src/main/java/com/google/cloud/tools/maven/CloudSdkMojo.java
TadCordle: Maybe, main reason they're separate right now is so I can mock them in tests.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
TadCordle: I think I can fix this in the next PR; it'll be easier to do when I have a ManagedCloudSdk I can get the downloaded path from.

TadCordle: Submitted an issue for this so I don't forget (#256)

PR #254, opened 2018-03-15
Add logic for cloud sdk installation/validation

TadCordle:


TadCordle:
@loosebazooka I actually tried going down that route a bit (see #253). One problem that arose when I started to implement `downloadCloudSdk` that way was having to add a `throws` clause to a ton of different methods. The cloud sdk builder is used all over the place, and it seemed cleanest to just throw a `MojoFailureException` to abort when the `ManagedCloudSdk` fails. Doing it in `execute()` seemed clean since it already throws `MojoFailureException`, rather than having to change a ton of different method headers.

I went with super.execute because it seemed like that was how the project handled goal dependencies ("deploy" calls "stage" using super.execute, for example). If you think putting it in CloudSdkAppEngineFactory is a better choice though I could reopen the old PR.

PR #261, opened 2018-03-20
Implement cloud sdk checker

### LINKED ISSUE #248: Implement cloud SDK validation

TadCordle:
Fixes #248

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkChecker.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkChecker.java
@@ -16,11 +16,49 @@
 
 package com.google.cloud.tools.maven;
 
-import org.apache.maven.plugin.logging.Log;
+import com.google.cloud.tools.appengine.cloudsdk.AppEngineJavaComponentsNotInstalledException;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkNotFoundException;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkOutOfDateException;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkVersionFileException;
+import com.google.common.annotations.VisibleForTesting;
 
 public class CloudSdkChecker {
 
-  public void checkCloudSdk(Log log) {
-    log.warn("Cloud SDK checker not yet implemented.");
+  private String version;
+
+  CloudSdkChecker(String version) {
+    this.version = version;
+  }
+
+  @VisibleForTesting
+  public void setVersion(String version) {
+    this.version = version;
+  }
+
+  /**
+   * Validates the cloud SDK installation
+   *
+   * @param cloudSdk CloudSdk with a configured sdk home directory
+   */
+  public void checkCloudSdk(CloudSdk cloudSdk) {
+    if (!version.equals(cloudSdk.getVersion().toString())) {
+      throw new RuntimeException(
+          "Specified Cloud SDK version ("
+              + version
+              + ") does not match installed version ("
+              + cloudSdk.getVersion()
+              + ").");
+    }
+
+    try {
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/maven/CloudSdkChecker.java
```diff
@@ -16,11 +16,49 @@
...
-  public void checkCloudSdk(Log log) {
-    log.warn("Cloud SDK checker not yet implemented.");
+  private String version;
+
+  CloudSdkChecker(String version) {
+    this.version = version;
+  }
+
+  @VisibleForTesting
+  public void setVersion(String version) {
```
loosebazooka: I wonder if we need this, give the constructor takes the same inputs?

src/main/java/com/google/cloud/tools/maven/CloudSdkOperationsFactory.java
```diff
@@ -23,19 +23,22 @@
...
+
+  CloudSdkOperationsFactory(CloudSdkMojo mojo) {
+    this.mojo = mojo;
   }
 
   /** Build a new ManagedCloudSdk from a given version */
   public ManagedCloudSdk newManagedSdk()
       throws UnsupportedOsException, BadCloudSdkVersionException {
-    if (Strings.isNullOrEmpty(version)) {
+    if (mojo == null || Strings.isNullOrEmpty(mojo.getCloudSdkVersion())) {
```
loosebazooka: can mojo really be null here?

src/main/java/com/google/cloud/tools/maven/CloudSdkOperationsFactory.java
```diff
@@ -23,19 +23,22 @@
...
 public class CloudSdkOperationsFactory {
-  private String version;
 
-  CloudSdkOperationsFactory(String version) {
-    this.version = version;
+  // At the factory's creation time, the sdk version isn't configured yet, so we can't pass that in
+  // by itself. So we pass in a reference to the CloudSdkMojo and get the version when we need it.
+  private final CloudSdkMojo mojo;
+
+  CloudSdkOperationsFactory(CloudSdkMojo mojo) {
```
loosebazooka: we're only using version in here aren't we?

TadCordle: The comment above explains it; mojo.getCloudSdkVersion() is still null at the time of the factory's instantiation, so I think we need a reference to the mojo, unless we want to instantiate the factory later on (which may make it trickier to test).

src/main/java/com/google/cloud/tools/maven/CloudSdkChecker.java
TadCordle: True. I was testing multiple versions and was injecting mocks, so I didn't want to re-instantiate the tested instance. But it looks like things have changed enough that I can remove this without affecting anything.

src/main/java/com/google/cloud/tools/maven/CloudSdkOperationsFactory.java
coollog: Could the version be passed into `newManagedSdk`/`newChecker`?

TadCordle: Maybe, would need to be passed into `newDownloader` as well since that uses `newChecker`. In that case though it would go back to what @loosebazooka and I were discussing earlier about needing tests to confirm the arguments are correctly used.

coollog: Hmm, yea it just seems a bit off due to the cyclical dependency where CloudSdkMojo uses CloudSdkAppengineFactory which uses CloudSdkMojo.

TadCordle: Pushed a change, let me know if it you think it needs tests.

PR #264, opened 2018-03-22
Remove appYamls parameter

### LINKED ISSUE #162: Delete deprecated appYamls parameter
Once we release the changes to use `<services>` instead of `<appYamls>` and it has been out there for a while so users can have enough time to deal with the deprecation, we should remove the `<appYamls>` parameter from the plugin.

TadCordle:
Fixes #162


src/main/java/com/google/cloud/tools/maven/RunMojo.java
```diff
@@ -283,7 +271,12 @@
   @Override
   public void execute() throws MojoExecutionException, MojoFailureException {
     SupportedDevServerVersion convertedVersion = convertDevserverVersionString();
-    handleAppYamlsDeprecation();
+    if (CollectionUtil.isNullOrEmpty(services)) {
+      Build build = mavenProject.getBuild();
+      services =
+          Collections.singletonList(
+              new File(build.getDirectory()).toPath().resolve(build.getFinalName()).toFile());
```
coollog: Not related to this change in particular, but from looking at this, it seems to be using the build artifact filename as a `service` input (which should be directories?) The javadoc for `getFinalName` also says that it `excludes the extension`, so it might not even be a valid file path. Also, services is a required property, so it seems that this should be a build failure rather than trying to find a default? I might just be not understanding how this works though.

@loosebazooka

loosebazooka: Yeah, it's a little strange, but since it's a list, if I recall correctly, we can't set a default on it directly using the `@Parameter` annotation.
I wonder if you can get the output directory directly from the maven project and that might be better? It does seem a little strange that we need to build this ourselves?

TadCordle: Closest thing I could find was `mavenProject.getBuild().getOutputDirectory`, but that returns a directory ending in `target/classes` instead of `target/<name>`.

TadCordle: @loosebazooka any idea what we should do here?

loosebazooka: It looks like the war plugin (https://maven.apache.org/plugins/maven-war-plugin/war-mojo.html) uses project.build.directory?

Perhaps we can do `mavenProject.getBuild().getDirectory()`, I haven't tested this. There might also be a way to grab the warPlugin.outputDirectory from the war plugin (and similarly for the jar plugin): Something like the way it's done in the jib plugin (https://github.com/google/jib/blob/ee67151df51c2e41196171baf27a31eabcf52c47/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProjectProperties.java#L79)?

loosebazooka: It does look like just using project.build.directory should allow the build to continue to work as before.

TadCordle: Isn't that what this code already does? build.getDirectory() + build.getFinalName()

loosebazooka: oops,  yeah. I guess I forgot the original context somewhere along the way. Perhaps we should just leave this as is for now?

PR #269, opened 2018-03-29
appengine-plugins-core 0.5.0 -> 0.5.2

TadCordle:


src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -60,7 +65,11 @@ private CloudSdkAppEngineFactory(
 
   @Override
   public AppEngineStandardStaging standardStaging() {
-    return cloudSdkFactory.standardStaging(defaultCloudSdkBuilder().build());
+    try {
+      return cloudSdkFactory.standardStaging(defaultCloudSdkBuilder().build());
+    } catch (CloudSdkNotFoundException ex) {
+      throw new RuntimeException(ex);
```
loosebazooka: I think we're going to need to restructure some stuff in plugins-core, so we don't have stuff this like :(

PR #279, opened 2018-04-05
Fix deployAll for flexible projects

### LINKED ISSUE #277: Fix appengineDeployAll task for flexible projects

TadCordle:
Fixes #277

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
@@ -40,29 +40,29 @@ public class DeployAllMojo extends DeployMojo {
 
     configureAppEngineDirectory();
 
-    // Make sure app.yaml is in the staging directory. Also check source directory for flexible
-    // projects.
-    if (!stagingDirectory.toPath().resolve("app.yaml").toFile().exists()
-        && (isStandardStaging()
-            || !appEngineDirectory.toPath().resolve("app.yaml").toFile().exists())) {
+    // Look for app.yaml
+    File appYaml = stagingDirectory.toPath().resolve("app.yaml").toFile();
+    if (!appYaml.exists() && !isStandardStaging()) {
+      appYaml = appEngineDirectory.toPath().resolve("app.yaml").toFile();
+    }
+    if (!appYaml.exists()) {
       throw new MojoExecutionException("Failed to deploy all: could not find app.yaml.");
     }
 
-    // Add app.yaml and config yamls to deployables
+    getLog().info("deployAll: Preparing to deploy app.yaml");
+    deployables.add(appYaml);
+
+    // Look for config yamls
     String[] configYamls = {"cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"};
     File configDir = isStandardStaging() ? stagingDirectory : appEngineDirectory;
-    addDeployable(stagingDirectory.toPath().resolve("app.yaml").toFile());
     for (String yamlName : configYamls) {
-      addDeployable(configDir.toPath().resolve(yamlName).toFile());
+      File yaml = configDir.toPath().resolve(yamlName).toFile();
+      if (yaml.exists()) {
+        getLog().info("deployAll: Preparing to deploy " + yaml.getName());
+        deployables.add(yaml);
+      }
     }
 
     super.execute();
-  }
-
-  private void addDeployable(File yaml) {
-    if (yaml.exists()) {
-      getLog().info("deployAll: Preparing to deploy " + yaml.getName());
-      deployables.add(yaml);
-    }
   }
 }
--- a/src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
@@ -17,6 +17,7 @@
 package com.google.cloud.tools.maven;
 
 import java.io.File;
+import java.nio.file.Path;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Execute;
@@ -42,7 +43,7 @@ public class DeployAllMojo extends DeployMojo {
 
     // Look for app.yaml
     File appYaml = stagingDirectory.toPath().resolve("app.yaml").toFile();
-    if (!appYaml.exists() && !isStandardStaging()) {
+    if (!isStandardStaging() && !appYaml.exists()) {
       appYaml = appEngineDirectory.toPath().resolve("app.yaml").toFile();
     }
     if (!appYaml.exists()) {
@@ -54,11 +55,11 @@ public class DeployAllMojo extends DeployMojo {
 
     // Look for config yamls
     String[] configYamls = {"cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"};
-    File configDir = isStandardStaging() ? stagingDirectory : appEngineDirectory;
+    Path configPath = isStandardStaging() ? stagingDirectory.toPath() : appEngineDirectory.toPath();
     for (String yamlName : configYamls) {
-      File yaml = configDir.toPath().resolve(yamlName).toFile();
+      File yaml = configPath.resolve(yamlName).toFile();
       if (yaml.exists()) {
-        getLog().info("deployAll: Preparing to deploy " + yaml.getName());
+        getLog().info("deployAll: Preparing to deploy " + yamlName);
         deployables.add(yaml);
       }
     }
--- a/src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
@@ -38,12 +38,19 @@ public class DeployAllMojo extends DeployMojo {
       deployables.clear();
     }
 
+    String[] configYamls = {"cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"};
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -247,7 +244,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     }
   }
 
-  protected boolean isStandardStaging() {
+  protected boolean isStandardStaging() throws MojoExecutionException {
+    if (appengineWebXml == null) {
+      appengineWebXml =
+          new AppEngineWebXml(new File(sourceDirectory.toString() + "/WEB-INF/appengine-web.xml"));
```
coollog: Should probably use path resolution to build the path rather than string concatenation.

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -268,7 +269,7 @@ protected void configureDockerfileDefaultLocation() {
    * <p>Called during {@link #execute()}, subclasses can override to provide a different value for
    * {@link #appEngineDirectory}.
    */
-  protected void configureAppEngineDirectory() {
+  protected void configureAppEngineDirectory() throws MojoExecutionException {
```
coollog: Is this `MojoExecutionException` used?

src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
```diff
@@ -46,6 +46,15 @@ public void execute() throws MojoExecutionException, MojoFailureException {
       if (yaml.exists()) {
         getLog().info("deployAll: Preparing to deploy " + yamlName);
         deployables.add(yaml);
+      } else if (!isStandardStaging()) {
```
coollog: I was thinking this logic could probably be organized as flexible vs. standard as the top-level condition, rather than having the standard yaml existence be the default check. In the case that the user was using flexible but had yamls in the standard locations, then it would be deploying possibly incorrect yamls. It might be fine, but it seems a bit off to me.

src/main/java/com/google/cloud/tools/maven/StageMojo.java
TadCordle: This method is overridden in AbstractSingleYamlDeployMojo and it's used there.

src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
TadCordle: As far as I understand, in flexible projects we do want to check for yamls to deploy in both the source directory and the staging directory, but I'm not sure. @loosebazooka?

src/main/java/com/google/cloud/tools/maven/StageMojo.java
coollog: Hmm, it seems that `isStandardStaging` should probably not set the `appengineWebXml` (like a method that should be a boolean getter should probably not be changing the state of the object).

coollog: Referring to `AbstractSingleYamlDeployMojo` using `isStandardStaging`, which is why it seems to throw the `MojoExecutionException`.

src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
TadCordle: Oh wait I see what you're saying I think

src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
```diff
@@ -38,17 +38,51 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+    String[] configYamls = {"cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"};
+    File appYaml = stagingDirectory.toPath().resolve("app.yaml").toFile();
+    if (appYaml.exists()) {
+      addDeployable(appYaml);
+    }
+
+    if (isStandardStaging()) {
+      // Check staging directory for app.yaml
+      if (!appYaml.exists()) {
+        throw new MojoExecutionException("Failed to deploy all: app.yaml not found.");
```
coollog: nit: since the exception here is the same as the exception for flexible, this could probably be split up into two stages, one for deploying app.yaml, and one for deploying the others, such that this code doesn't have to be duplicated.

src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
```diff
@@ -38,17 +38,51 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-      "app.yaml", "cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"
-    };
-    for (String yamlName : validYamls) {
-      File yaml = stagingDirectory.toPath().resolve(yamlName).toFile();
-      if (yaml.exists()) {
-        getLog().info("deployAll: Preparing to deploy " + yamlName);
-        deployables.add(yaml);
+    String[] configYamls = {"cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"};
+    File appYaml = stagingDirectory.toPath().resolve("app.yaml").toFile();
+    if (appYaml.exists()) {
```
coollog: nit: Looks like `addDeployable` could even include the existence check

src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
```diff
@@ -38,26 +38,48 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
     }
 
-    String[] validYamls = {
-      "app.yaml", "cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"
-    };
-    for (String yamlName : validYamls) {
-      File yaml = stagingDirectory.toPath().resolve(yamlName).toFile();
+    if (isStandardStaging()) {
+      // Check staging directory for all yamls
+      String[] validYamls = {
```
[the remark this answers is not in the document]
TadCordle: Looks like I fixed these right before you commented on them.

src/main/java/com/google/cloud/tools/maven/DeployAllMojo.java
```diff
@@ -38,17 +38,51 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+        if (appYaml.exists()) {
+          addDeployable(appYaml);
+        } else {
+          throw new MojoExecutionException("Failed to deploy all: app.yaml not found.");
+        }
+      }
+
+      // Look appengine directory for config yamls
+      configureAppEngineDirectory();
+      for (String yamlName : configYamls) {
```
coollog: nit: This could probably also be combined with the loop for standard, just changing the directory it resolves for.

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -216,15 +215,15 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-
     configureAppEngineDirectory();
 
     if (isStandardStaging()) {
       getLog().info("Detected App Engine standard environment application.");
 
       // force runtime to 'java' for compat projects using Java version >1.7
+      AppEngineWebXml appengineWebXml =
+          new AppEngineWebXml(
+              new File(sourceDirectory.toPath().resolve("WEB-INF/appengine-web.xml").toString()));
```
loosebazooka: `.resolve("WEB-INF").resolve("appengine-web.xml")`

but also, maybe I missed some older comments, why did we need to move this around?

TadCordle: I wanted to call `isStandardStaging()` in `DeployAllMojo`, but the original implementation used `appengineWebXml`, which was just a private field that was set in `StageMojo.execute(). So whenever I tried to call `isStandardStaging()`, `appengineWebXml` was null and I'd get an error. So I changed `isStandardStaging()` to not rely on the field, realized `appengineWebXml` wasn't used anywhere else, and reduced its scope.

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -248,7 +252,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
   }
 
   protected boolean isStandardStaging() {
-    return appengineWebXml.exists();
+    return Files.exists(sourceDirectory.toPath().resolve("WEB-INF/appengine-web.xml"));
```
loosebazooka: probably want to resolve resolve here too.

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -216,15 +215,20 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
     getLog().info("Staging the application to: " + stagingDirectory);
 
-    appengineWebXml =
-        new AppEngineWebXml(new File(sourceDirectory.toString() + "/WEB-INF/appengine-web.xml"));
-
     configureAppEngineDirectory();
```
loosebazooka: feels like you could just resolve appengienWeb.xml here and do an if(exists) to start this loop?

loosebazooka: unless isStandardStaging is used somewhere else.

TadCordle: I use isStandardStaging in the deploy all task, which is why I changed this in the first place.

TadCordle: Also in `AbstractSingleYamlDeployMojo`.

PR #284, opened 2018-04-12
Log devappserver output to file for appengine:start

### LINKED ISSUE #10: For gcp-app:start write devappserver output to a file

TadCordle:
Fixes #10

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
@@ -38,6 +38,7 @@ import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListen
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import org.apache.maven.plugin.logging.Log;
 
@@ -238,7 +239,7 @@ public class CloudSdkAppEngineFactory {
   /**
    * Default output listener that copies output to the Maven Mojo logger with a 'GCLOUD: ' prefix.
    */
-  protected static class DefaultProcessOutputLineListener implements ProcessOutputLineListener {
+  static class DefaultProcessOutputLineListener implements ProcessOutputLineListener {
 
     private final Log log;
 
@@ -253,13 +254,13 @@ public class CloudSdkAppEngineFactory {
   }
 
   /** A listener that redirects process output to a file. */
-  protected static class FileOutputLineListener implements ProcessOutputLineListener {
+  static class FileOutputLineListener implements ProcessOutputLineListener {
 
     private final PrintStream logFilePrinter;
 
-    public FileOutputLineListener(File logFile) {
+    FileOutputLineListener(File logFile) {
       try {
-        logFilePrinter = new PrintStream(logFile, "UTF-8");
+        logFilePrinter = new PrintStream(logFile, StandardCharsets.UTF_8.name());
       } catch (IOException ex) {
         throw new RuntimeException(ex);
       }
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
@@ -40,6 +40,7 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import org.apache.maven.plugin.logging.Log;
 
 /** Factory for App Engine dependencies. */
@@ -158,12 +159,14 @@ public class CloudSdkAppEngineFactory {
    */
   public AppEngineDevServer devServerRunAsync(
       int startSuccessTimeout, SupportedDevServerVersion version) {
-    File logFile =
-        new File(
-            mojo.mavenProject.getBuild().getDirectory() + "/dev-appserver-out/dev_appserver.out");
-    if (!logFile.getParentFile().exists() && !logFile.getParentFile().mkdirs()) {
+    File logDir =
+        Paths.get(mojo.mavenProject.getBuild().getDirectory())
+            .resolve("dev-appserver-out")
+            .toFile();
+    if (!logDir.exists() && !logDir.mkdirs()) {
       throw new RuntimeException("Failed to create dev-appserver logging directory.");
     }
+    File logFile = new File(logDir, "dev_appserver.out");
     FileOutputLineListener fileListener = new FileOutputLineListener(logFile);
     mojo.getLog().info("Dev App Server output written to : " + logFile);
 
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
@@ -39,6 +39,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.apache.maven.plugin.logging.Log;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -154,8 +157,16 @@ private AppEngineDevServer createDevServerForVersion(
    */
   public AppEngineDevServer devServerRunAsync(
       int startSuccessTimeout, SupportedDevServerVersion version) {
+    File logFile = new File(mojo.mavenProject.getBuild().getDirectory(), "dev_appserver.out");
```
coollog: Should the log output directory be configurable?

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -241,6 +252,25 @@ public void onOutputLine(String line) {
     }
   }
 
+  /** A listener that redirects process output to a file. */
+  protected static class FileOutputLineListener implements ProcessOutputLineListener {
```
coollog: Is the protected visibility necessary?

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -241,6 +252,25 @@ public void onOutputLine(String line) {
...
+
+    private final PrintStream logFilePrinter;
+
+    public FileOutputLineListener(File logFile) {
+      try {
+        logFilePrinter = new PrintStream(logFile, "UTF-8");
```
coollog: I believe `PrintStream` should be closed after use.

coollog: For the "UTF-8" constant, there's `StandardCharsets.UTF_8.name()`.

src/test/java/com/google/cloud/tools/maven/it/RunAsyncMojoIntegrationTest.java
```diff
@@ -63,6 +66,20 @@ private void test(String name, SupportedDevServerVersion version)
...
+          Paths.get(
+                  "target",
+                  "test-classes",
+                  "projects",
+                  "standard-project",
+                  "target",
+                  "dev_appserver.out")
+              .toFile();
+      assertTrue(expectedLog.exists());
+      String devAppServerOutput = FileUtils.readFileToString(expectedLog, "UTF-8");
```
coollog: There's `Files.readAllBytes` to operate on a `Path`.

src/test/java/com/google/cloud/tools/maven/it/RunAsyncMojoIntegrationTest.java
```diff
@@ -63,6 +66,20 @@ private void test(String name, SupportedDevServerVersion version)
       String urlContent = UrlUtils.getUrlContentWithRetries(getServerUrl(), 60000, 1000);
       assertThat(urlContent, containsString("Hello from the App Engine Standard project."));
       assertThat(urlContent, containsString("TEST_VAR=testVariableValue"));
+
+      File expectedLog =
+          Paths.get(
+                  "target",
```
coollog: Is this explicit pathing the only way to get to this file?

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
TadCordle: It doesn't look like it is in the gradle plugin.

coollog: Ah okay, it looks like in the gradle plugin, it puts the out file in a dev-appserver-out directory.

https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/156/files#diff-f80e036458fbefa4397a6bedcdc45897R45

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
TadCordle: I was mimicking the `DefaultProcessOutputLineListener` definition, but yeah it looks like both can be package private.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
TadCordle: I'm not sure where I'd be able to close this (the gradle plugin uses the same class but doesn't close it either).

src/test/java/com/google/cloud/tools/maven/it/RunAsyncMojoIntegrationTest.java
TadCordle: Not sure, is there a way to get the build directory from within the integration tests @loosebazooka?

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
coollog: Perhaps by attaching an exit listener?

src/test/java/com/google/cloud/tools/maven/it/RunAsyncMojoIntegrationTest.java
```diff
@@ -67,17 +67,16 @@ private void test(String name, SupportedDevServerVersion version)
...
-      assertTrue(expectedLog.exists());
-      String devAppServerOutput = FileUtils.readFileToString(expectedLog, "UTF-8");
+              "target",
+              "test-classes",
+              "projects",
+              "standard-project",
+              "target",
+              "dev_appserver.out");
+      assertTrue(expectedLog.toFile().exists());
+      String devAppServerOutput = new String(Files.readAllBytes(expectedLog));
```
coollog: String construction should take the charset as well.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
coollog: It also looks like PrintWriter wrapping a BufferedWriter might be preferred for appending strings to a file over PrintStream, but this should be fine since it's how the gradle plugin does it.

https://stackoverflow.com/questions/11372546/printstream-vs-printwriter

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -158,7 +158,12 @@ private AppEngineDevServer createDevServerForVersion(
    */
   public AppEngineDevServer devServerRunAsync(
       int startSuccessTimeout, SupportedDevServerVersion version) {
-    File logFile = new File(mojo.mavenProject.getBuild().getDirectory(), "dev_appserver.out");
+    File logFile =
+        new File(
+            mojo.mavenProject.getBuild().getDirectory() + "/dev-appserver-out/dev_appserver.out");
```
coollog: Should probably use path resolution here.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -158,7 +158,12 @@ private AppEngineDevServer createDevServerForVersion(
...
+    if (!logFile.getParentFile().exists() && !logFile.getParentFile().mkdirs()) {
```
coollog: The dev-appserver-out should probably be defined as itself instead of using the `getParentFile`. Also, it might be clearer to put this check before resolving logFile.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -258,9 +263,16 @@ public void onOutputLine(String line) {
 
     private final PrintStream logFilePrinter;
 
-    FileOutputLineListener(File logFile) {
+    FileOutputLineListener(final File logFile) {
       try {
         logFilePrinter = new PrintStream(logFile, StandardCharsets.UTF_8.name());
+        Runtime.getRuntime()
```
coollog: This is probably fine, but I was referring to adding an `exitListener` that would close this `PrintStream` on like L170. Would require getting the PrintStream from this object though.

src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -154,8 +159,23 @@ private AppEngineDevServer createDevServerForVersion(
    */
   public AppEngineDevServer devServerRunAsync(
       int startSuccessTimeout, SupportedDevServerVersion version) {
+    File logDir =
+        Paths.get(mojo.mavenProject.getBuild().getDirectory())
+            .resolve("dev-appserver-out")
+            .toFile();
+    if (!logDir.exists() && !logDir.mkdirs()) {
```
coollog: nit: since this is all new code, we generally prefer to go with the `nio.Path` API rather than the ([kinda deprecated](https://stackoverflow.com/questions/30547589/why-isnt-java-io-file-deprecated)) `io.File`. For example, since you are working with a Path already, the exists and mkdirs can be done with `Files.exists` and `Files.createDirectories`.

src/test/java/com/google/cloud/tools/maven/it/RunAsyncMojoIntegrationTest.java
```diff
@@ -63,6 +67,21 @@ private void test(String name, SupportedDevServerVersion version)
...
+              "target",
+              "test-classes",
+              "projects",
+              "standard-project",
+              "target",
+              "dev-appserver-out",
+              "dev_appserver.out");
+      assertTrue(Files.exists(expectedLog));
+      String devAppServerOutput =
+          new String(Files.readAllBytes(expectedLog), StandardCharsets.UTF_8.name());
```
coollog: Oh, here the String constructor can take the Charsets as a parameter, so `.name` doesn't need to be called. Only older APIs like `PrintStream` need the string name of the charset.

PR #291, opened 2018-04-17
Use appengine-web.xml as backup config instead of gcloud global state

### LINKED ISSUE #285: Read application/version from pom.xml + appengine-web.xml instead of pom.xml + gcloud global state
1. If "deploy.read.appengine.web.xml" system property is set, make sure project id is only set in appengine-web.xml and not build config, otherwise error.
2. If "deploy.read.appengine.web.xml" is not set, then make sure project id is set in build config, otherwise error.

TadCordle:
Fixes #285

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/AbstractDeployMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/AbstractDeployMojo.java
@@ -117,10 +117,6 @@ public abstract class AbstractDeployMojo extends StageMojo implements DeployConf
   }
 
   void updatePropertiesFromAppEngineWebXml() throws IOException, SAXException, AppEngineException {
-    if (!isStandardStaging()) {
-      return;
-    }
-
     AppEngineDescriptor appengineWebXmlDoc =
         AppEngineDescriptor.parse(
             new FileInputStream(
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/tools/maven/AbstractDeployMojo.java
```diff
@@ -110,4 +115,53 @@ public String getVersion() {
   public String getProject() {
     return project;
   }
+
+  void updatePropertiesFromAppEngineWebXml() throws IOException, SAXException, AppEngineException {
+    if (!isStandardStaging()) {
```
loosebazooka: I wonder if we can restructure this so we can share code without using the parent class.

AppEngineDeployer {
  static fromConfig(....)
  void stage();
  void deploy();
  void deployAll();
  void deployIndexes();
  ...
} 

// returns either a standard or flexible deployer, each of which does the right thing for their respective environment for the deployAll, deploy, deployIndexes or whatever tasks.
// reuse this deployer everywhere. (deploy, all, indexes, cron, etc)

This moves the code that makes decisions out of the parent classes like `AbstractSingleYamlDeployMojo#configureAppEngineDirectory` or whatever. However it is not a small rewrite. So if you're willing, lets try to do it, or we can push this first and then schedule a rewrite after this change.

TadCordle: I think a rewrite is necessary, because right now the gcloud config update stuff only applies to deploy and deployAll, since they're the only goals that extend `AbstractDeployMojo`. I can fix this in the short term by either:
1. copying project ID verification code over to `AbstractSingleYamlDeployMojo` (easiest, but kinda messy), or
2. by just merging `AbstractDeployMojo` and `AbstractSingleYamlDeployMojo` together, since it looks like they share some properties anyway.

I think I'd prefer to do one of the short term fixes first (probably the second one) to get all the functionality in there, and then go for the bigger rewrite afterwards.

loosebazooka: Well maybe if we're not releasing this until 2.0, we can just ignore the deploySingleYaml for now. I can't tell if option 2 really works and if it does, if that temporary rewrite is worth it. I think we'd probably rather just move towards some kind of Deployer object and solve the deploySingleYaml problem at the same time.

I think we would still want to do this before 2.0, it would be really nice to get rid of this whole "execute" chain before the release. Whaddya think?

TadCordle: Makes sense, I was thinking even if it's a temporary rewrite, we'd still be deleting a file, which would make the bigger rewrite a bit easier. But yeah, it could cause unnecessary complications.

I think we should merge this first and then go for it.

PR #297, opened 2018-04-23
Refactor to avoid using super.execute()

### LINKED ISSUE #292: Refactor to share code without using parent class
Original comment:

>I wonder if we can restructure this so we can share code without using the parent class.
>
>`AppEngineDeployer {
  static fromConfig(....)
  void stage();
  void deploy();
  void deployAll();
  void deployIndexes();
  ...
} `
// returns either a standard or flexible deployer, each of which does the right thing for their respective environment for the deployAll, deploy, deployIndexes or whatever tasks.
// reuse this deployer everywhere. (deploy, all, indexes, cron, etc)
>
>This moves the code that makes decisions out of the parent classes like AbstractSingleYamlDeployMojo#configureAppEngineDirectory or whatever. However it is not a small rewrite. So if you're willing, lets try to do it, or we can push this first and then schedule a rewrite after this change.

TadCordle:
Fixes #292.

Most stage/deploy functionality is now pulled out into implementations of a new `AppEngineDeployer` interface so code can be shared between the mojos without forcing them to execute each other. The parent class hierarchy still exists for sharing configuration parameters.


src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -195,96 +186,28 @@
...
   public void execute() throws MojoExecutionException, MojoFailureException {
-    if (!"war".equals(getPackaging()) && !"jar".equals(getPackaging())) {
-      // https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/85
-      getLog().info("Staging is only executed for war and jar modules.");
-      return;
-    }
+    AppEngineDeployer.Factory.newDeployer(this).stage();
+  }
+
+  protected void clearStagingDirectory() throws MojoFailureException, MojoExecutionException {
```
loosebazooka: Presumably this code should be moved out of here too.

src/main/java/com/google/cloud/tools/maven/AppEngineFlexibleDeployer.java
```diff
@@ -0,0 +1,168 @@
...
+            .toPath()
+            .resolve("src")
+            .resolve("main")
+            .resolve("appengine")
+            .toFile();
+  }
+
+  @Override
+  public void deploy() {
+    AbstractDeployMojo deployMojo = (AbstractDeployMojo) stageMojo;
```
loosebazooka: Hrm... I guess this is something we have deal with: this kind of typecasting guess? Maybe you need to have AppEngineStager and AppEngineDeployer be two different types?

src/main/java/com/google/cloud/tools/maven/DeployIndexMojo.java
```diff
@@ -16,25 +16,22 @@
...
-      DeployProjectConfigurationConfiguration configuration) {
-    try {
-      appEngineDeployment.deployIndex(this);
-    } catch (AppEngineException ex) {
-      throw new RuntimeException(ex);
-    }
+  @Override
+  public void execute() throws MojoExecutionException, MojoFailureException {
+    AppEngineDeployer deployer = AppEngineDeployer.Factory.newDeployer(this);
+    deployer.configureAppEngineDirectory();
```
loosebazooka: So we don't actually need to stage on flex executions of this task do we?

I wonder if `stage` should just be an internal call of `deployer.deployIndex`, similarly `configureAppEngineDirectory` can be internal to deployer?

TadCordle: The reason I left `configureAppEngineDirectory` separate is because `deploy` and `deployAll` don't call it before `stage` (they use src/main/appengine regardless of standard or flexible).

I also originally kept `stage` by itself because testing was difficult (it kept deleting my test files for the `deployAll` test when it cleared the staging directory). I could split things out into more methods to make testing deploy functionality without staging easier, but then I'd end up with double the number of deployment methods in these classes, which could be messy.

loosebazooka: Oh right... you need to call it before stage.

Could we do something like this?
```
class Stager {}
class Deployer {
  Stager stager
}
```

TadCordle: Ah yeah, that's a good idea. Solves the problem from your other comment as well.

PR #306, opened 2018-05-09
Read project deployment config from all sources

### LINKED ISSUE #305: Read project deployment config from all sources.
Introduces some keyworks for 'project' that allow the user to configure what they want as their source of truth during deployment

Allow project/version config to be read from

1. pom.xml (this is default)
2. appengine-web.xml : `<project>APPENGINE_CONFIG</project>`
3. gcloud global state: `<project>'GCLOUD_CONFIG'</project>`

`APPENGINE_CONFIG`/`GCLOUD_CONFIG` are not valid google cloud project
names and are reconfigured by the plugin.

Flexible projects do not allow the user to configure `APPENGINE_CONFIG`

TadCordle:
* ~~Adds AbstractAppEngineDeployer class to clean up duplicate code between the standard and flex deployers.~~
* Moves `deploy` and `deploy<Config>` functionality from deployers to respectives goals, since the code is exactly the same between standard and flex deployments.
* Adds new logic mimicking gradle plugin to read project/version from build config, appengine-web.xml, or gcloud state.
* Fixes #305.


CHANGELOG.md
```diff
@@ -13,10 +13,11 @@ All notable changes to this project will be documented in this file.
 * Upgrade App Engine Plugins Core dependency to 0.5.2.
 * `cloudSdkPath` has been replaced with `cloudSdkHome`. ([#257](../../issues/257))
 * Remove deprecated `appYamls` parameter. ([#162](../../issues/162))
-* `project` and `version` are no longer pulled from the global gcloud state. `project` must be configured
-in either pom.xml using the `<project>` tag or in appengine-web.xml using the `<application>` tag. 
-`version` may be left unconfigured to be generated by gcloud, or it may be configured in your pom.xml or 
-appengine-web.xml. To read values from appengine-web.xml, set the system property `deploy.read.appengine.web.xml=true`. ([#285](../../issues/285))
+* `project` and `version` are no longer pulled from the global gcloud state by default. `project`
```
loosebazooka: I think this might need to made more maven specific

`<project>APPENGINE_CONIFG</project>` vs `project = "APPENGINE_CONFIG"`

src/main/java/com/google/cloud/tools/maven/AppEngineFlexibleDeployer.java
```diff
@@ -79,64 +63,47 @@ public void deployAll() throws MojoExecutionException, MojoFailureException {
...
-    try {
-      deployMojo.getAppEngineFactory().deployment().deployIndex(deployMojo);
-    } catch (AppEngineException ex) {
-      throw new RuntimeException(ex);
+  void updateGcloudProperties() throws MojoExecutionException {
+    if (deployMojo.project == null
+        || deployMojo.project.trim().isEmpty()
+        || deployMojo.project.equals(APPENGINE_CONFIG)) {
+      throw new MojoExecutionException(
+          "Deployment project must be defined or configured to read from system state\n"
```
loosebazooka: Same thing here with the maven specific config?

```
Deploy project must be defined or set to read from system state, by setting <project> in your plugin configuration
1. Set <project>my-project-name</project>
2. Set <project>GCLOUD_CONFIG</project> to blah blah blah
3. ...
```

src/main/java/com/google/cloud/tools/maven/AbstractAppEngineDeployer.java
```diff
@@ -0,0 +1,106 @@
...
+      updateGcloudProperties();
+      deployMojo.getAppEngineFactory().deployment().deployCron(deployMojo);
+    } catch (AppEngineException ex) {
+      throw new RuntimeException(ex);
+    }
+  }
+
+  @Override
+  public void deployDispatch() throws MojoFailureException, MojoExecutionException {
+    stager.configureAppEngineDirectory();
```
loosebazooka: I think the case against making these common is that flex deployments of configuration files do not need to stage. The Flex version of these would actually skip that step.

And with only deploy being common that point, we might as well just put that in the Flex/Standard deployer.

updateGcloudProperties will just be an internal implementation of each Deployer?

TadCordle: I see, yeah that makes sense.

loosebazooka:
gonna go ahead and close this one

PR #307, opened 2018-05-14
Add in new project/version config, move some code around

loosebazooka:
@TadCordle take a look? This is kind of a duplicate of #306 but I went and changed some of the underlying mojo structure.


src/main/java/com/google/cloud/tools/maven/AppEngineStager.java
```diff
@@ -16,25 +16,20 @@
...
+          ? new AppEngineStandardStager()
+          : new AppEngineFlexibleStager();
     }
   }
 
-  void stage() throws MojoExecutionException, MojoFailureException;
+  void stage(StageMojo stageMojo) throws MojoExecutionException, MojoFailureException;
 
-  void configureAppEngineDirectory();
+  void setAppEngineDirectory(StageMojo stageMojo);
```
TadCordle: Maybe this method should be named something different, since it's not a typical setter.

src/main/java/com/google/cloud/tools/maven/AppEngineStandardDeployer.java
```diff
@@ -19,177 +19,191 @@
...
+    stager.stage(deployMojo);
+    deployMojo.setDeployables(ImmutableList.of(deployMojo.getStagingDirectory()));
 
     try {
-      updatePropertiesFromAppEngineWebXml();
       deployMojo.getAppEngineFactory().deployment().deploy(deployMojo);
-    } catch (AppEngineException | SAXException | IOException ex) {
-      throw new MojoFailureException(ex.getMessage(), ex);
+    } catch (AppEngineException ex) {
+      throw new MojoExecutionException("Standard Application deployment failed", ex);
```
TadCordle: "Application" should probably be lowercase.

src/main/java/com/google/cloud/tools/maven/AppEngineStager.java
loosebazooka: I took another look at this, turns out we don't need this at all. appengineDirectory is a flex only parameter and we can force it to be `src/main/appengine`.

src/main/java/com/google/cloud/tools/maven/StageMojo.java
```diff
@@ -181,14 +175,17 @@
 
   @Override
   public void execute() throws MojoExecutionException, MojoFailureException {
```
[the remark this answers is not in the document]
loosebazooka: Yeah it's definitely super weird. Let me play around with compound config, see what we can do.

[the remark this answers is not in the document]
loosebazooka: I realize, I misunderstood what you were saying. And yeah we can probably limit the scope of what deploys have access to, I'll do that.

src/main/java/com/google/cloud/tools/maven/AppEngineStandardDeployer.java
```diff
@@ -19,177 +19,190 @@
...
 
+    deployMojo.setDeployables(standardDeployables.build());
+
     try {
-      updatePropertiesFromAppEngineWebXml();
       deployMojo.getAppEngineFactory().deployment().deploy(deployMojo);
-    } catch (AppEngineException | SAXException | IOException ex) {
-      throw new RuntimeException(ex);
+    } catch (AppEngineException ex) {
+      throw new MojoExecutionException("Failed to Deploy", ex);
```
TadCordle: Missed a few of these (`deployAll`, `deployCron`, `deployDispatch`, `deployDos` in `AppEngineStandardDeployer`)

loosebazooka: ah.. I'll go fix those.

Issue #312, opened 2018-06-10
Allow running AppEngine for another project using devserver.services


matt-blanchette:
The previous Maven plugin for AppEngine had a configuration property "appDir" 
to allow deploying another project.

The current plugin only allows deploying the current Maven project 
and will fail when the `WEB-INF/appengine-web.xml` is not found: 
`Dev App Server does not support App Engine Flexible Environment applications.`

It appears the property `devserver.services` should allow the same functionality,
if it was used in RunMojo.isStandardEnvironmentApp()

matt-blanchette:
I've tested this fix in RunMojo:
> protected boolean isStandardEnvironmentApp() {
      return new File(services.get(0), "/WEB-INF/appengine-web.xml").exists();
  }

The RunMojoTest would need `src/main/appengine` changed to a valid standard env path like `src/test/resources/projects/standard-project/src/main/webapp`

matt-blanchette:
Added pull request #313

loosebazooka:
Can you explain how you are using this? Is this is a workaround for using a parent pom with multiple modules?

matt-blanchette:
We have a separate UI automation project to deploy the main application using start/stop goals as integration test.

matt-blanchette:
This can be closed, now that #313 was merged

PR #313, opened 2018-06-11
Fixed run mojo to use services param in standard env app check #312

matt-blanchette:


src/test/java/com/google/cloud/tools/maven/RunMojoTest.java
```diff
@@ -50,6 +51,11 @@
 @RunWith(JUnitParamsRunner.class)
 public class RunMojoTest extends AbstractDevServerTest {
 
+  private static final String STANDARD_PROJECT_WEBAPP =
```
loosebazooka: just need to add a negative test here, where it fails (one standard app, one non-standard app)

matt-blanchette: 👍

matt-blanchette:
Any way to re-trigger these checks?

loosebazooka:
looks good, thanks for your contributions!

PR #317, opened 2018-06-22
Instantiate CloudSdk later

loosebazooka:
mojo may not be initialized with values if done at mojo
construction time

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
+++ b/src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
@@ -48,7 +48,6 @@ import org.apache.maven.plugin.logging.Log;
 /** Factory for App Engine dependencies. */
 public class CloudSdkAppEngineFactory {
 
-  private final CloudSdk cloudSdk;
   private final CloudSdkMojo mojo;
 
   /** Supported dev app server versions. */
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/maven/CloudSdkAppEngineFactory.java
```diff
@@ -140,10 +129,14 @@ public GenRepoInfoFile genRepoInfoFile() {
     return getGcloud().newGenRepoInfo(newDefaultProcessHandler());
   }
 
+  private CloudSdk getCloudSdk() {
+    return defaultCloudSdk(mojo, new CloudSdkOperationsFactory());
```
coollog: We prob don't want to create a new `CloudSdk` each this this is called right?

loosebazooka: yeah, I guess we could singleton it, but I think each Mojo already has it's own instance of this Factory anyway. So unless we rewrite this whole thing as some kind of Singleton service that instantiated once, it shouldn't matter.

coollog: I was thinking that if `mojo.getCloudSdkHome` was null, it would be downloading the CloudSDK every time `getCloudSdk` was called (L143).

loosebazooka: Download should only run conditionally, this is handled further down in the code? Lemme find it.

loosebazooka: Ah CloudSdkDownloader only conditionally downloads it.

loosebazooka: I think there some space to clean up this whole interaction. will do in a followup.

coollog: It looks like the checking on L157 still happens right?

PR #322, opened 2018-08-30
project -> projectId, proejctId on run

loosebazooka:
- project -> projectId
- run now supports projectId
- Update to plugins core 0.6.6
- Fix some deprecated guava Files calls

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/ConfigReader.java
+++ b/src/main/java/com/google/cloud/tools/maven/ConfigReader.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2018 Google LLC. All Rights Reserved.
+ *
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/cloud/tools/maven/ConfigReader.java
```diff
@@ -0,0 +1,111 @@
...
+    private final File appengineWebXml;
+
+    private AppEngineWebXmlConfig(File appengineWebXml) {
+      this.appengineWebXml = appengineWebXml;
+    }
+
+    /** Return "application" tag from appengine-web.xml or error out if could not read. */
+    public String getProject() {
+      try {
+        AppEngineDescriptor appEngineDescriptor =
```
coollog: ditto combining this and `getVersion`?

PR #327, opened 2018-08-31
2.0.0 rc2

loosebazooka:


pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-maven-plugin</artifactId>
-  <version>2.0.0-SNAPSHOT</version>
+  <version>2.0.0-SNAPSHOT-SNAPSHOT</version>
```
coollog: Oh this looks like something went wrong.

loosebazooka: yeah it looks like maven-release autoinjects the `SNAPSHOT`, I'm going to have to make that clear in the release script

PR #328, opened 2018-08-31
only append SNAPSHOT if necessary

loosebazooka:


loosebazooka:
gonna ignore the travis builds for now.

scripts/prepare_release.sh
```diff
@@ -67,7 +67,10 @@ git tag v${VERSION}
 
 # Updates the pom.xml with the next snapshot version.
 # For example, when releasing 1.5.7, the next snapshot version would be 1.5.8-SNAPSHOT.
-NEXT_SNAPSHOT=${NEXT_VERSION}-SNAPSHOT
```
coollog: Shouldn't we just have `<post-release-version>` not include `-SNAPSHOT` when passing it in?

loosebazooka: That's not clear, this still handles both behaviors correctly though.

PR #334, opened 2018-10-09
Make transition smoother to projectId

loosebazooka:
Reintroduces the parameter (removed in rc2), to provide an effective transition from 1.3.X GA users.

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/AbstractDeployMojo.java
+++ b/src/main/java/com/google/cloud/tools/maven/AbstractDeployMojo.java
@@ -63,7 +63,12 @@ public abstract class AbstractDeployMojo extends AbstractStageMojo
   @Parameter(alias = "deploy.version", property = "app.deploy.version")
   private String version;
 
-  /** The Google Cloud Platform project name to use for this invocation. */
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/tools/maven/AbstractDeployMojo.java
```diff
@@ -63,7 +63,12 @@
   @Parameter(alias = "deploy.version", property = "app.deploy.version")
   private String version;
 
-  /** The Google Cloud Platform project name to use for this invocation. */
+  /** The Google Cloud Platform project Id to use for this invocation */
+  @Deprecated
```
chanseokoh: It seems a bit unusual that you "deprecate" a parameter that did not exist. I guess this is basically for forcing a crash when the user tries to set this "new" parameter? I wonder what happens currently if the user sets this parameter.

loosebazooka: Hah, yeah it used to exist (it was replaced with projectId in the previous release). The appengine team asked for a smoother transition in the config. So doing this.

PR #346, opened 2018-12-12
Update integration tests

loosebazooka:
- add project/version params
- update to java 8


loosebazooka:
Should be the last one before we setup to do the next RC release of the plugins

PR #354, opened 2019-01-03
Upgrade Mockito and plugins / clean up warnings

chanseokoh:
In preparation for building on JDK 11.

Upgrades Mockito and some plugins. Some tests were failing after upgrading Mockito because of unnecessary stubbing. I removed them but used `Mockito.lenient()` in one place to suppress the error, which seemed meaningful have the stubbing.

This will allow building on JDK 11 almost successfully, but unfortunately, the dev appserver v2 doesn't seem to run on JVM 11, so integration tests with dev appserver v2 fail. Not sure if recent Cloud SDK versions ship with Java 11 compatible dev appserver v2 and our tests were merely pulling in an old Cloud SDK, but I haven't gone that far to figure out if it is the case.

### DIFF
--- a/src/test/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkDownloaderTest.java
+++ b/src/test/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkDownloaderTest.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.maven.cloudsdk;
 
+import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -38,7 +39,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
-import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnitRunner;
 
 @RunWith(MockitoJUnitRunner.class)
 public class CloudSdkDownloaderTest {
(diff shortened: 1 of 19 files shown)


src/test/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkDownloaderTest.java
```diff
@@ -85,7 +85,7 @@ public void testDownloadCloudSdk_installAppEngine()
   public void testDownloadCloudSdk_ignoreAppEngineComponent()
       throws ManagedSdkVerificationException, ManagedSdkVersionMismatchException {
     when(managedCloudSdk.isInstalled()).thenReturn(true);
-    when(managedCloudSdk.hasComponent(SdkComponent.APP_ENGINE_JAVA)).thenReturn(false);
+    lenient().when(managedCloudSdk.hasComponent(SdkComponent.APP_ENGINE_JAVA)).thenReturn(false);
```
loosebazooka: What does this do?

chanseokoh: Mockito fails if there is any stubbing that's not being used. In this test, no one called `hasComponent(...)`, so Mockito wants us to remove this `when()`. To make Mockito tolerate this, we can use `Mockito.lenient()` instead of removing `when()`. Removing it works too, especially knowing that the default return value for `boolean` is false anyway.

loosebazooka: ah, I see, then maybe we can just remove this line.

PR #355, opened 2019-01-04
Remove Cobertura

chanseokoh:
Cobertura can't handle Java 8 code, and `./mvnw -U cobertura:cobertura` is failing. Cobertura has been dormant for years regarding Java 8 support and seems almost dead.

### DIFF
--- a/kokoro/continuous.bat
+++ b/kokoro/continuous.bat
@@ -7,7 +7,7 @@ cd github/app-maven-plugin
 call gcloud.cmd components update --quiet
 call gcloud.cmd components install app-engine-java --quiet
 
-call mvnw.cmd clean install cobertura:cobertura -B -U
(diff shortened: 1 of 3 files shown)


chanseokoh:
Jacoco sounds good, which our CT4E also uses. However, I admit I am totally lost with the code coverage business. This repo is already doing something with codecov.io, and it's confusing to me what all these (Codecov, JaCoCo, Cobertura) achieve and how.

loosebazooka:
I think it's just whoever is generating the reports, I assume we will get no reports if no one is generating them?

chanseokoh:
Ah, we already have JaCoCo: https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/2.0.0-rc4/pom.xml

PR #361, opened 2019-01-10
do not delete staging directory by default

loosebazooka:
This allows users to inject arbitrary files into staging using maven `copy-resources` `copy-dependencies` goals


chanseokoh:
So, are we going to document somewhere something like "the default staging directory is target/appengine-staging where gcloud will be invoked to deploy, and if needed, you can copy extra files (for example) using `copy-resources`"?

loosebazooka:
Yeah, basically anyone who is using those goals will have to configure those correctly to reference the appengine staging directory (either the default or whatever they have configured)

loosebazooka:
I should also add that we will still allow the use to add in files using `extraFilesDirectories`

loosebazooka:
@ludoch how do you feel about this solution?

PR #369, opened 2019-02-12
Update to plugins core 0.7.4

loosebazooka:
- also fix scope of maven-verifier dependency to "test"


loosebazooka:
NOTE that this includes a change from plugins core that reverts java 11 standard-compat support.

loosebazooka:
Java11 standard compat was experimental and ended as just an experiment

PR #378, opened 2019-04-08
Remove v2-alpha devappserver support

loosebazooka:
Also adds better error messages for app.yaml-based-project users trying to use this.


loosebazooka:
actually gotta clean up one more file.

chanseokoh:
Tests fail on MacOS, consistently. Interestingly they are OK on Linux and Windows.

loosebazooka:
Actually this appears to be a new failure related to `random`

```
[INFO] GCLOUD: java.security.ProviderException: setSeed() failed
[INFO] GCLOUD: 	at sun.security.provider.NativePRNG$RandomIO.implSetSeed(NativePRNG.java:472)
```

loosebazooka:
@chanseokoh I might have to admin override and push this, while we wait on the MacOS build to be fixed. Can you do a review?

src/test/java/com/google/cloud/tools/maven/run/RunnerTest.java
```diff
@@ -97,13 +95,9 @@ private void setUpAppEngineWebXml() throws IOException {
   }
 
   @Test
-  @Parameters({"1,V1", "2-alpha,V2ALPHA"})
-  public void testRun(String version, SupportedDevServerVersion mockVersion)
-      throws MojoExecutionException, IOException, AppEngineException {
+  public void testRun() throws MojoExecutionException, IOException, AppEngineException {
```
chanseokoh: Probably you can remove `@RunWith(JUnitParamsRunner.class)`.

src/test/java/com/google/cloud/tools/maven/it/RunAsyncMojoIntegrationTest.java
```diff
@@ -46,22 +44,12 @@ public void initPorts() throws IOException {
...
-  }
-
-  private void test(String name, SupportedDevServerVersion version)
+  private void testRunAsync(String name)
       throws IOException, VerificationException, InterruptedException {
+    String testName = "testRunAsync";
     try {
-      Verifier verifier = createVerifier(name, version);
+      Verifier verifier = createVerifier(testName);
+      verifier.setSystemProperty("app.devserver.port", Integer.toString(serverPort));
```
chanseokoh: This is done in `createVerifier()` too. Maybe just remove `createVerifier()` and just do `new StandardVerifier(testName)`, or move much of code here to `createVerifier()` if duplicated repeatedly in other tests?

loosebazooka: Yeah since this is used twice, I've left the extra `createVerifier` in and cleaned it up as you suggested.

src/test/java/com/google/cloud/tools/maven/run/RunnerTest.java
loosebazooka: Switched to Mockito Junit Runner

src/test/java/com/google/cloud/tools/maven/it/RunAsyncMojoIntegrationTest.java
```diff
@@ -38,30 +37,17 @@
...
-  }
-
-  private void test(String name, SupportedDevServerVersion version)
-      throws IOException, VerificationException, InterruptedException {
+  public void testRunAsync() throws IOException, VerificationException, InterruptedException {
+    String testName = "testRunAsync";
     try {
-      Verifier verifier = createVerifier(name, version);
+      Verifier verifier = createVerifier(testName);
       verifier.setSystemProperty("app.devserver.startSuccessTimeout", "60");
```
chanseokoh: I just find it weird you set `app.devserver.port` inside `createVerifier()` but set `app.devserver.startSuccessTimeout` outside it.

loosebazooka: Haha, I think it's because `stop` doesn't use it (further down in `finally`)? I dunno, I didn't write this  ¯\_(ツ)_/¯

chanseokoh: Alright, that's fair.

Unfortunately, some tests are failing. Other than that LGTM.

loosebazooka: weird different versions of dev appserver print different started strings. Using the same one as we do in gradle for these checks.

loosebazooka:
NOTE: These macos build errors are related to a `/dev/urandom` failure on our macos build test machines. Intentional admin override here.

PR #380, opened 2019-04-10
Kokoro macos temporary fix - use newer JDK

loosebazooka:
internal reference: b/130225695

### DIFF
--- a/kokoro/continuous_mac.sh
+++ b/kokoro/continuous_mac.sh
@@ -8,7 +8,7 @@ set -x
 gcloud components update
 gcloud components install app-engine-java
 
-# Use adopt openjdk 8u202 until kokoro updates it's macos images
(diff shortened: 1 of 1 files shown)


loosebazooka:
The build is actually passing, some weird aborted build stuff is going on, let me retrigger this.

kokoro/continuous_mac.sh
```diff
@@ -8,6 +8,12 @@ set -x
 gcloud components update
 gcloud components install app-engine-java
 
+# Use adopt openjdk 8u202 until kokoro updates it's macos images
```
chanseokoh: Add the bug number? The comment will become questionable if looked at a couple years later.

loosebazooka: Makes sense. Will do.

PR #381, opened 2019-04-12
Set up OpenJDK 11 build

### LINKED ISSUE #372: Set up JDK 11 build on Travis once dev appserver 2 is gone
The reason I couldn't set up a JDK 11 build in #354 was because of dev appserver 2.

chanseokoh:
Fixes #372.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
 sudo: false
 language: java
-jdk:
-  - oraclejdk8
+matrix:
+  include:
+    - name: Oracle JDK 8
+      jdk: oraclejdk8
+    - name: OpenJDK 11
+      jdk: openjdk11
 cache:
   directories:
-   - $HOME/.m2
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,11 +1,15 @@
...
+  include:
+    - name: Oracle JDK 8
+      jdk: oraclejdk8
+    - name: OpenJDK 11
+      jdk: openjdk11
 cache:
   directories:
-   - $HOME/.m2
-   - $HOME/google-cloud-sdk
+    - $HOME/.m2
```
loosebazooka: Is our whole file formatted incorrectly? This seems to be different than that cr change you made?

loosebazooka: actually looks like people in general just want it to be consistent with itself

chanseokoh: I also got confused. Maybe the formatting in the CL is an internal style.

PR #383, opened 2019-04-16
2.0.0 rc6

loosebazooka:


chanseokoh:
Shouldn't we Increase the SNAPSHOT version to `2.0.1-SNAPSHOT`? Nothing is being changed.

loosebazooka:
@chanseok, yeah we're going from `2.0.0-SNAPSHOT` -> `2.0.0-rc6` -> `2.0.0-SNAPSHOT`, but since we still plan on doing a `2.0.0` next week, moving to `2.0.1` doesn't make sense here.

PR #385, opened 2019-04-17
Remove v2-alpha cleanup stragglers

loosebazooka:


src/main/java/com/google/cloud/tools/maven/run/AbstractRunMojo.java
```diff
@@ -51,8 +51,6 @@
   /**
    * Restart instances automatically when files relevant to their module are changed. (default:
    * True)
-   *
-   * <p><i>Supported only for devserver version 2-alpha.</i>
```
chanseokoh: If the comment is true, we should remove this parameter?

loosebazooka: The comment is false. We can set automatic restart on devappserver 1.

PR #394, opened 2019-07-16
Add gcloudMode config option for beta deploy

### LINKED ISSUE #374: gcloud beta app deploy
I can't find a way to make it run "gcloud beta app deploy" instead of "gcloud app deploy" when running "appengine:deploy". Is there any? I want to use "network: session_affinity: true" and it seems that using "gcloud beta app deploy" is the only way to do it.

loosebazooka:
fixes #374 

### usage
```
<configuration>
  ...
  <gcloudMode>beta</gcloudMode>
  ...
</configuration>
```

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkAppEngineFactory.java
+++ b/src/main/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkAppEngineFactory.java
@@ -34,6 +34,7 @@ import com.google.cloud.tools.appengine.operations.cloudsdk.process.LegacyProces
 import com.google.cloud.tools.appengine.operations.cloudsdk.process.NonZeroExceptionExitListener;
 import com.google.cloud.tools.appengine.operations.cloudsdk.process.ProcessHandler;
 import com.google.cloud.tools.appengine.operations.cloudsdk.process.ProcessOutputLineListener;
+import com.google.cloud.tools.managedcloudsdk.components.SdkComponent;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/cloud/tools/maven/deploy/AppDeployer.java
```diff
@@ -176,6 +176,7 @@ public void deployQueue() throws MojoExecutionException {
     DeployConfiguration buildDeployConfiguration(List<Path> deployables) {
       return DeployConfiguration.builder(deployables)
           .bucket(deployMojo.getBucket())
+          .gcloudMode(deployMojo.getGcloudMode())
```
chanseokoh: If the configuration is missing, will `getGcloudMode()` return null or an empty string? In either case, will it be safe?

src/main/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkAppEngineFactory.java
```diff
@@ -130,12 +133,16 @@ static CloudSdk buildCloudSdk(
         return cloudSdk;
       } else {
         // we need to use a managed cloud sdk
+        List<SdkComponent> requiredComponents = new ArrayList<>();
+        if (requiresAppEngineComponents) {
+          requiredComponents.add(SdkComponent.APP_ENGINE_JAVA);
```
chanseokoh: There's no prod code that requests to install `SdkComponent.BETA`. Is the API change necessary?

src/main/java/com/google/cloud/tools/maven/deploy/AppDeployer.java
loosebazooka: yeah it's safe to pass null, all config params can be null.

src/main/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkAppEngineFactory.java
loosebazooka: I think it's okay, I like this pattern more, rather than passing around some boolean all over the place.

PR #396, opened 2019-08-05
update travis jdk8 build

loosebazooka:


.travis.yml
```diff
@@ -2,8 +2,9 @@ sudo: false
 language: java
 matrix:
   include:
-    - name: Oracle JDK 8
+    - name: Oracle JDK 8 on Trusty
```
loosebazooka: see jib: https://github.com/GoogleContainerTools/jib/pull/1874

chanseokoh: I think we should prefer `openjdk8` if it works. In the Jib case, `openjdk8` has a Javadoc-related issue.

loosebazooka: I think we ran into the same issue with javadoc, that's why we use openjdk, but I don't remember exactly.

loosebazooka: let me try.

loosebazooka:
closed in favor of #397

PR #401, opened 2019-09-16
Rearrage docs, add faq for thin jars

loosebazooka:
You can see the rendered docs on the branch:
* readme: https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/rearrange-docs/README.md
* user_guide: https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/rearrange-docs/USER_GUIDE.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ You can now run commands like `mvn package appengine:deploy` in the root folder 
 
 Please see the [USER GUIDE](USER_GUIDE.md) for a full list of supported goals and configuration
 options.
-* [for app.yaml based projects](USER_GUIDE.md#app-engine-appengine-webxml-based-projects)
-* [for appengine-web.xml based projects](USER_GUIDE.md#app-engine-appyaml-based-project)
+* [USER\_GUIDE for `app.yaml` based projects](USER_GUIDE.md#app-engine-appyaml-based-projects)
+* [USER\_GUIDE for `appengine-web.xml` based projects](USER_GUIDE.md#app-engine-appengine-webxml-based-project)
 
 # Reference Documentation
 
--- a/README.md
+++ b/README.md
@@ -5,10 +5,6 @@
 
 This Maven plugin provides goals to build and deploy Google App Engine applications.
 
-| 2.0.0 is Live |
-| :------------------------------ |
-| 2.0.0 has been published. The behavior of the appengine-maven-plugin has changed since v1.+; please see the [CHANGELOG](CHANGELOG.md) for a full list of changes and an updated [USER GUIDE](USER_GUIDE.md) for details. If you are having trouble using or updating your plugin, please file a [new issue](https://github.com/GoogleCloudPlatform/app-maven-plugin/issues).|
-
 # Requirements
 
 [Maven](http://maven.apache.org/) is required to build and run the plugin.
--- a/README.md
+++ b/README.md
@@ -9,29 +9,12 @@ This Maven plugin provides goals to build and deploy Google App Engine applicati
 | :------------------------------ |
 | 2.0.0 has been published. The behavior of the appengine-maven-plugin has changed since v1.+; please see the [CHANGELOG](CHANGELOG.md) for a full list of changes and an updated [USER GUIDE](USER_GUIDE.md) for details. If you are having trouble using or updating your plugin, please file a [new issue](https://github.com/GoogleCloudPlatform/app-maven-plugin/issues).|
 
-# Reference Documentation
-
-App Engine Standard Environment:
-* [Using Apache Maven and the App Engine Plugin (standard environment)](https://cloud.google.com/appengine/docs/java/tools/using-maven)
-* [App Engine Maven Plugin Goals and Parameters (standard environment)](https://cloud.google.com/appengine/docs/java/tools/maven-reference)
-
-App Engine Flexible Environment:
-* [Using Apache Maven and the App Engine Plugin (flexible environment)](https://cloud.google.com/appengine/docs/flexible/java/using-maven)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -47,38 +30,19 @@ In your Maven App Engine Java app, add the following plugin to your pom.xml:
...
-appengine:deployQueue|Deploy queue configuration.
-appengine:deployAll|Deploy the application with all available valid configuration.
-appengine:run|Runs the App Engine local development server. *(App Engine Standard Only)*
-appengine:start|Starts running the App Engine devserver asynchronously and then returns to the command line. When this goal runs, the behavior is the same as the run goal except that Maven continues processing goals and exits after the server is up and running. *(App Engine Standard Only)*
-appengine:stop|Stops a running App Engine web development server. *(App Engine Standard Only)*
-appengine:genRepoInfoFile|Generates source context files for use by Stackdriver Debugger.
-appengine:help|Displays help information on the plugin. Use `mvn appengine:help -Ddetail=true -Dgoal=[goal]` for detailed goal documentation.
+Please see the [USER GUIDE](USER_GUIDE.md) for a full list of supported goals and configuration
+options.
+* [for app.yaml based projects](USER_GUIDE.md#app-engine-appengine-webxml-based-projects)
```
TadCordle: These two links are reversed

loosebazooka: whoops, good catch

USER_GUIDE.md
```diff
@@ -368,4 +368,41 @@ The plugin defaults to `appengine-web.xml` based deployment if your project cont
 file. If your project also has an `src/main/appengine/app.yaml` file and you wish to use that, you may temporarily move the
 `appengine-web.xml` file to a different location before deploying.
 
+### How do I deploy an `app.yaml` based thin jar to appengine?
```
TadCordle: App Engine?

USER_GUIDE.md
```diff
@@ -368,4 +368,41 @@ The plugin defaults to `appengine-web.xml` based deployment if your project cont
...
+
+You will probably need to stage the application to include it's dependencies. You might use something like the
```
TadCordle: ~~it's~~ its

loosebazooka: I rewrote this whole thing, can you take another look?

README.md
```diff
@@ -9,29 +9,12 @@ This Maven plugin provides goals to build and deploy Google App Engine applicati
 | :------------------------------ |
 | 2.0.0 has been published. The behavior of the appengine-maven-plugin has changed since v1.+; please see the [CHANGELOG](CHANGELOG.md) for a full list of changes and an updated [USER GUIDE](USER_GUIDE.md) for details. If you are having trouble using or updating your plugin, please file a [new issue](https://github.com/GoogleCloudPlatform/app-maven-plugin/issues).|
```
TadCordle: Should this section say 2.1.0 now, or is it fine to leave it as 2.0.0 since it's focused on v1 -> v2 migration?

USER_GUIDE.md
```diff
@@ -368,4 +368,51 @@ The plugin defaults to `appengine-web.xml` based deployment if your project cont
...
+  <configuration>
+    <!-- your config -->
+    <extraFilesDirectories>
+       <extraFilesDirectory>${project.build.directory}/dependencies</extraFilesDirectory>
+    </extraFilesDirectories>
+  </configuration>
+</plugin>
+```
+
+Then when you run `mvn package appengine:deploy` the dependencies are copied and included as part of the deployment
```
TadCordle: Missing period at the end

README.md
loosebazooka: I think I can take this down and make it a section below. It's already been like 5 months since 2.0.0

loosebazooka: actually, I'm just going to remove it and do nothing.

USER_GUIDE.md
```diff
@@ -368,4 +368,51 @@ The plugin defaults to `appengine-web.xml` based deployment if your project cont
...
+```
+
+Then when you run `mvn package appengine:deploy` the dependencies are copied and included as part of the deployment.
+
+Since the default entrypoint assumes a fat jar, you must define a custom entrypoint to run this application. In this
+example we might have an `app.yaml` that looks like:
+
+```
+runtime: java11
+entrypoint: java -Xmx64m -cp "*" com.example.MyMainClass
```
chanseokoh: Is `-Xmx64m` recommended for App Engine? If not, remove it?

USER_GUIDE.md
```diff
@@ -368,4 +368,51 @@ The plugin defaults to `appengine-web.xml` based deployment if your project cont
...
+  <configuration>
+    <!-- your config -->
+    <extraFilesDirectories>
+      <extraFilesDirectory>${project.build.directory}/dependencies</extraFilesDirectory>
+    </extraFilesDirectories>
+  </configuration>
+</plugin>
+```
+
+Then when you run `mvn package appengine:deploy` the dependencies are copied and included as part of the deployment.
```
chanseokoh: I could be wrong about this, but just in case:

I think a lot of frameworks nowadays will by default create a fat JAR, so maybe it'd be nice to explain a bit for those cases to deploy an app correctly. For example, Spring Boot will generate two JARs (fat one and the original one), and isn't it that this command will copy the fat JAR to the staging area (or both)?

BTW, there seems to be a working hack the user and @ludoch used to remove a fat JAR.

USER_GUIDE.md
loosebazooka: That's a good point, it was in the example on the appengine docs, so I used it.

USER_GUIDE.md
loosebazooka: Yes, I think I might lean towards thinking that that's on them. For springboot, in maven, I believe some of the configurations leads to it overriding the default final artifact as the springboot fat jar (but I could be wrong).

PR #405, opened 2019-10-22
Upgrade plugins-core to 0.8.1

loosebazooka:
Adds extended support for thin jar deployments, automatically copying
jars referenced in Class-Path in the source artifact to the staging
directory.

@ludoch this is all we would need to support this feature in 2.2.0


loosebazooka:
Travis builds will not pass until tomorrow (when maven central sync happens).
kokoro-macos is acting crazy, going to try and see if that can be fixed.

loosebazooka:
I'm going to submit this despite travis failures.

PR #407, opened 2019-10-24
Update README.md and CHANGELOG.md

loosebazooka:


loosebazooka:
Travis seems to be still failing on being unable to resolve artifacts from the google maven mirror: https://maven-central.storage-download.googleapis.com/repos/central/data/com/google/cloud/tools/appengine-plugins-core/maven-metadata.xml (which is slightly out of sync with maven central)

Issue #421, opened 2020-06-10
Failed to run devappserver: NoSuchFileException: appengine-web.xml


RayDiazVega:
Hello

I'm trying to run a spring boot maven flex app with App Engine, I'm using `java 1.8.0_251`, `Apache Maven 3.6.3`, `Google Cloud SDK [296.0.0]`, `app-engine-java: [1.9.80]`,  `IntelliJ IDEA 2020.1.2`, this is the appengine plugin in my pom:

```
<plugin>
  <artifactId>appengine-maven-plugin</artifactId>
  <configuration>
    <deploy.projectId>my-project</deploy.projectId>
    <deploy.version>GCLOUD_CONFIG</deploy.version>
  </configuration>
  <groupId>com.google.cloud.tools</groupId>
  <version>2.2.0</version>
</plugin>
```

And this is part of my app.yaml:

```
service: reports
env: flex
runtime: java
runtime_config:
  jdk: openjdk8
```

So when I execute my project with the command `mvn package appengine:run`, I get the following error:

`Failed to execute goal com.google.cloud.tools:appengine-maven-plugin:2.2.0:run (default-cli) on project reports: Failed to run devappserver: java.nio.file.NoSuchFileException: C:\Users\ACER1\Documents\proyects\reports\target\reports-0.0.1-SNAPSHOT\WEB-INF\appengine-web.xml`

When I generate the project in [start.spring.io](https://start.spring.io/), no appengine-web.xml file was generated. Do I need to add this file on my own for the `mvn package appengine:run` command works?

Thanks for your time.

loosebazooka:
`appengine:run` is not available for flex projects.

chanseokoh:
The flex and standard are two different environments, and there are a lot of factors you should consider when choosing an environment that's right for you. Only because the standard env has `appengine:run` doesn't sound like a logical choice to switch to standard. However, your app is just a hello-world app, so I think it's still good to use standard. You app is a Spring Boot project, so the runtime you can use on standard is [Java 11 Standard Environment](https://cloud.google.com/appengine/docs/standard/java11/runtime). (Unless you do something very weird, Java 8 apps will have no problem on Java 11 JVM.)

> Or is there some other way to run my flex project locally?

[...]

PR #423, opened 2020-06-20
Handle empty service config better

loosebazooka:
fixes the error message for non appengine-web.xml projects when not specifying a service.

https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/421#issuecomment-642873454

### DIFF
--- a/src/main/java/com/google/cloud/tools/maven/run/Runner.java
+++ b/src/main/java/com/google/cloud/tools/maven/run/Runner.java
@@ -89,7 +89,8 @@ public class Runner {
   List<Path> processServices() throws MojoExecutionException {
     List<Path> services = runMojo.getServices();
 
-    Preconditions.checkState(services != null && !services.isEmpty(), "No services found to run");
+    Preconditions.checkState(services != null, "'services' is null") ;
+    Preconditions.checkState(!services.isEmpty(), "'services' is empty");
 
     // verify all are appengine-web.xml based applications
     for (Path service : services) {
--- a/src/main/java/com/google/cloud/tools/maven/run/Runner.java
+++ b/src/main/java/com/google/cloud/tools/maven/run/Runner.java
@@ -20,12 +20,10 @@ import com.google.cloud.tools.appengine.AppEngineException;
 import com.google.cloud.tools.appengine.configuration.RunConfiguration;
 import com.google.cloud.tools.maven.cloudsdk.ConfigReader;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.Collections;
 import java.util.List;
-import org.apache.maven.model.Build;
 import org.apache.maven.plugin.MojoExecutionException;
 
 public class Runner {
@@ -82,11 +80,6 @@ public class Runner {
     runMojo.getLog().info("Use the 'mvn appengine:stop' command to stop the server.");
   }
 
-  private Path getAppDir() {
-    Build build = runMojo.getMavenProject().getBuild();
-    return Paths.get(build.getDirectory()).resolve(build.getFinalName());
-  }
-
   static final String NON_STANDARD_APPLICATION_ERROR =
       "\nCould not find appengine-web.xml all services, perhaps you need to run "
           + "'mvn package appengine:run/start'."
@@ -94,11 +87,9 @@ public class Runner {
 
   @VisibleForTesting
   List<Path> processServices() throws MojoExecutionException {
-    Path appDir = getAppDir();
     List<Path> services = runMojo.getServices();
-    if (services == null || services.isEmpty()) {
-      return Collections.singletonList(appDir);
-    }
+
+    Preconditions.checkState(services != null && !services.isEmpty(), "No services found to run");
 
     // verify all are appengine-web.xml based applications
     for (Path service : services) {
--- a/src/main/java/com/google/cloud/tools/maven/run/Runner.java
+++ b/src/main/java/com/google/cloud/tools/maven/run/Runner.java
@@ -89,7 +89,7 @@ public class Runner {
   List<Path> processServices() throws MojoExecutionException {
     List<Path> services = runMojo.getServices();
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/tools/maven/run/Runner.java
```diff
@@ -82,23 +80,16 @@ public void runAsync(int startSuccessTimeout) throws MojoExecutionException {
...
 
   @VisibleForTesting
   List<Path> processServices() throws MojoExecutionException {
-    Path appDir = getAppDir();
     List<Path> services = runMojo.getServices();
-    if (services == null || services.isEmpty()) {
-      return Collections.singletonList(appDir);
-    }
+
+    Preconditions.checkState(services != null && !services.isEmpty(), "No services found to run");
```
chanseokoh, line 92: I prefer two check conditions with two different error messages, so that in case of a failure at runtime, we can easily figure out which condition was violated.

src/test/java/com/google/cloud/tools/maven/run/AbstractRunMojoTest.java
```diff
@@ -0,0 +1,3 @@
+package com.google.cloud.tools.maven.run;
+
+public class AbstractRunMojoTest {}
```
chanseokoh, line 3: Remove it.

src/test/java/com/google/cloud/tools/maven/run/RunnerTest.java
```diff
@@ -120,25 +111,29 @@ public void testRunAsync() throws MojoExecutionException, IOException, AppEngine
   public void testProcessServices_singleService() throws MojoExecutionException {
     List<Path> userConfiguredServices = ImmutableList.of(STANDARD_PROJECT_WEBAPP);
     when(runMojo.getServices()).thenReturn(userConfiguredServices);
-    List<Path> processedServices = testRunner.processServices();
-    Assert.assertEquals(userConfiguredServices, processedServices);
+    testRunner.processServices();
+    // no exception pass
```
chanseokoh, line 115: Can we not preserve the `assertEquals` test? Seems meaningful to check the return value.

src/test/java/com/google/cloud/tools/maven/run/RunnerTest.java
```diff
@@ -120,25 +111,29 @@ public void testRunAsync() throws MojoExecutionException, IOException, AppEngine
...
 
   @Test
   public void testProcessServices_multipleProjects() throws MojoExecutionException {
     List<Path> userConfiguredServices =
         ImmutableList.of(STANDARD_PROJECT_WEBAPP, STANDARD_PROJECT_WEBAPP2);
     when(runMojo.getServices()).thenReturn(userConfiguredServices);
-    List<Path> processedServices = testRunner.processServices();
-    Assert.assertEquals(userConfiguredServices, processedServices);
+    testRunner.processServices();
+    // no exception pass
```
chanseokoh, line 136: ditto

src/main/java/com/google/cloud/tools/maven/run/Runner.java
loosebazooka: Technically this is unreachable. `getServices()` cannot (and should not) return `null or empty`. This more of an assertion than anything.

src/test/java/com/google/cloud/tools/maven/run/RunnerTest.java
loosebazooka: Yeah, I went back and forth on that, this does less "processing" now. But I can change it back, makes sense.

src/test/java/com/google/cloud/tools/maven/run/AbstractRunMojoTest.java
loosebazooka: oops, forgot to fill out this test class

src/main/java/com/google/cloud/tools/maven/run/Runner.java
loosebazooka: changed this anyway.

src/test/java/com/google/cloud/tools/maven/run/AbstractRunMojoTest.java
```diff
@@ -0,0 +1,68 @@
+package com.google.cloud.tools.maven.run;
```
chanseokoh, line 1: copyright

src/test/java/com/google/cloud/tools/maven/run/AbstractRunMojoTest.java
```diff
@@ -0,0 +1,68 @@
...
+        throws NoSuchFieldException, IllegalAccessException {
+      Field servicesField = AbstractRunMojo.class.getDeclaredField("services");
+      servicesField.setAccessible(true);
+      servicesField.set(this, services);
+    }
+  }
+
+  @Mock MavenProject mavenProject;
+
+  @InjectMocks private final AbstractRunMojoImpl testMojo = new AbstractRunMojoImpl();
```
chanseokoh, line 38: Does `@InjectMocks` do something? (I have never used `@InjectMocks`, so I don't know what it is for.) Looks like you do manual injection with `inejctServices()`?

src/test/java/com/google/cloud/tools/maven/run/AbstractRunMojoTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public void injectServices(List<File> services)
+        throws NoSuchFieldException, IllegalAccessException {
+      Field servicesField = AbstractRunMojo.class.getDeclaredField("services");
+      servicesField.setAccessible(true);
+      servicesField.set(this, services);
+    }
+  }
+
+  @Mock MavenProject mavenProject;
```
chanseokoh, line 36: `private`

src/test/java/com/google/cloud/tools/maven/run/AbstractRunMojoTest.java
loosebazooka: Inject mocks, injects the `mavenProject` into the object at mock construction time. It doesn't work for Services because those change per test. (I could probably also do this with a package private setter)

PR #429, opened 2020-09-10
add verbosity option

loosebazooka:


src/main/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkMojo.java
```diff
@@ -42,6 +42,9 @@
   @Parameter(property = "serviceAccountKeyFile", required = false)
   private File serviceAccountKeyFile;
 
+  /** Optional parameter to configure the gcloud logging verbosity level. */
+  @Parameter private String verbosity;
```
chanseokoh, line 46: Is it unnecessary to have
```
@Parameter(property = "verbosity", required = false)
```
like for the `serviceAccountKeyFile` field?

loosebazooka: property is for the command line option `-Dverbosity=info` or something, which I'm hesitant to add, maybe it could be `gcloud.verbosity` or something? `verbosity` seems like it could clash with a bunch of things

loosebazooka: I will add `gcloud.verbosity`

chanseokoh (approved):

Issue #428, opened 2020-08-18
Misleading log message when app.yaml and appengine-web.xml file are missing


peterjot:
If `app.yaml` and `appengine-webl.xml` file are missing then we get "Detected App Engine app.yaml based application." log message. I think we could log another message in such case, because this one is misleading.

https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/master/src/main/java/com/google/cloud/tools/maven/deploy/Deployer.java#L44

jasinapj:
Maybe we could throw an exception if the xml and yaml configuration were not found? Right now the yaml stager is used in that case

loosebazooka:
Makes sense, and user contributions are always appreciated.

cxorm:
Could I take this issue if you don't mind : )

meltsufin:
Sure.

cxorm:
Thanks @meltsufin , I'm working on it.

elefeint:
`p4` means you don't have a deadline to finish the task :)

PR #451, opened 2021-09-11
Throw exception when app.yaml doesn't exist

cxorm:
Fixes GoogleCloudPlatform/appengine-plugins#990

## What changes were proposed in this pull request ?

The [Stager Interface](https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/master/src/main/java/com/google/cloud/tools/maven/stage/Stager.java#L39#L42) only check `appengine-web.xml`, and it would work when both `app.yaml` and `appengine-web.xml` are missing.

This patch verified the existence of `app.yaml` in [`AppYamlStager#stage()`](https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/master/src/main/java/com/google/cloud/tools/maven/stage/AppYamlStager.java#L41). 

**PS.** Let me steal the idea of snippet in [`AppDeployer`](https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/master/src/main/java/com/google/cloud/tools/maven/deploy/AppDeployer.java#L68#L72)  : ) 

## How was this patch tested ?

Just touch new unit test. (Feel free to correct me if I miss something.)

### DIFF
--- a/src/test/java/com/google/cloud/tools/maven/stage/AppYamlStagerTest.java
+++ b/src/test/java/com/google/cloud/tools/maven/stage/AppYamlStagerTest.java
@@ -23,10 +23,13 @@ import com.google.cloud.tools.appengine.operations.AppYamlProjectStaging;
 import com.google.cloud.tools.maven.cloudsdk.CloudSdkAppEngineFactory;
 import com.google.cloud.tools.maven.stage.AppYamlStager.ConfigBuilder;
 import junitparams.JUnitParamsRunner;
+import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.project.MavenProject;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
+import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.mockito.InjectMocks;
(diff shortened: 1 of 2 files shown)


cxorm:
@googlebot I signed it!

src/test/java/com/google/cloud/tools/maven/stage/AppYamlStagerTest.java
```diff
@@ -58,4 +61,16 @@ public void configureStageMojo() {
...
     when(stagingConfiguration.getStagingDirectory()).thenReturn(tempFolder.getRoot().toPath());
   }
+
+  @Test
+  public void noAppYaml() {
+    when(stageMojo.getStagingDirectory()).thenReturn(tempFolder.getRoot().toPath());
+    testStager = new AppYamlStager(stageMojo, configBuilder);
+
+    try {
+      testStager.stage();
```
chanseokoh, line 71: Should have `Assert.fail()`.

chanseokoh:
Actually, reading #428, I think it's better to check the existence of either `app.yaml` or `appengine-web.xml` [here](https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/9d9d6ec3dce9891713d12d8dd8f52e06e1f2c19a/src/main/java/com/google/cloud/tools/maven/deploy/Deployer.java#L44) and throw an error message, rather than assuming it's an `app.yaml`-based project and have the `AppYamlStager` throw an error?

chanseokoh:
Oh, maybe the deployer is not involved when only staging? But I think it's still better to check early for the existence of an either file that can show a clear error message than falling through the `app.yaml`-based workflow.

PR #454, opened 2021-11-30
Test–do not merge: Build and test using Java 17

chanseokoh:
Let me see what happens.


chanseokoh:
At least it needs to bump the major version of maven-surefire-plugin (currently only 3.0 milestone versions available).

elefeint:
Superceded by #455

PR #455, opened 2022-03-25
Working version with Java 17

elefeint:
All the plugins are working. I upgraded findbugs to a version somewhat lower than the latest because the latest comes with deeper immutability analysis that's more involved to fix (and not clear it's worth fixing for stable, self-contained code).

The integration test using AppEngine local devserver had to pass the JPMS add-opens parameteres, since devserver does some access modification of JDK classes through reflection. Since JPMS is available only starting at Java 9, I split the sample application appengine plugin into two profiles activated by java versions.
Also, a timezone option needed to be passed to work around an NPE in devserver (it's been fixed in devserver since, but not released yet).

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -39,6 +39,18 @@ For appengine-web.xml based projects the plugin exposes the following goals :
 | `start`           | Start the application in the background. |
 | `stop`            | Stop a running application. |
 
+NOTE: To run locally on Java 17, you'll need to add the following plugin configuration:
+
+```
+<configuration>
+  <jvmFlags>
+    <jvmFlag>--add-opens=java.base/java.net=ALL-UNNAMED</jvmFlag>
+    <jvmFlag>--add-opens=java.base/sun.net.www.protocol.http=ALL-UNNAMED</jvmFlag>
+    <jvmFlag>--add-opens=java.base/sun.net.www.protocol.https=ALL-UNNAMED</jvmFlag>
+  </jvmFlags>
+</plugin>
+```
+
 #### Deployment
 
 | Goal             | Description |
--- a/src/test/resources/projects/standard-project/pom.xml
+++ b/src/test/resources/projects/standard-project/pom.xml
@@ -66,12 +66,9 @@
                 <jvmFlag>-Djava.security.egd=file:/dev/urandom</jvmFlag>
 
                 <!-- Illegal reflective access disallowed since Java 16 -->
-                <jvmFlag>--add-opens</jvmFlag>
-                <jvmFlag>java.base/java.net=ALL-UNNAMED</jvmFlag>
-                <jvmFlag>--add-opens</jvmFlag>
-                <jvmFlag>java.base/sun.net.www.protocol.http=ALL-UNNAMED</jvmFlag>
-                <jvmFlag>--add-opens</jvmFlag>
-                <jvmFlag>java.base/sun.net.www.protocol.https=ALL-UNNAMED</jvmFlag>
+                <jvmFlag>--add-opens=java.base/java.net=ALL-UNNAMED</jvmFlag>
+                <jvmFlag>--add-opens=java.base/sun.net.www.protocol.http=ALL-UNNAMED</jvmFlag>
+                <jvmFlag>--add-opens=java.base/sun.net.www.protocol.https=ALL-UNNAMED</jvmFlag>
 
                 <jvmFlag>-Duser.timezone='GMT-5'</jvmFlag>
 
--- a/src/test/resources/projects/standard-project/pom.xml
+++ b/src/test/resources/projects/standard-project/pom.xml
@@ -27,7 +27,16 @@
           <jvmFlags>
             <!-- required for kokoro gcp_ubuntu b/37217450 -->
             <jvmFlag>-Djava.security.egd=file:/dev/urandom</jvmFlag>
+            <jvmFlag>--add-opens</jvmFlag>
+            <jvmFlag>java.base/java.net=ALL-UNNAMED</jvmFlag>
+            <jvmFlag>--add-opens</jvmFlag>
+            <jvmFlag>java.base/sun.net.www.protocol.http=ALL-UNNAMED</jvmFlag>
+            <jvmFlag>--add-opens</jvmFlag>
+            <jvmFlag>java.base/sun.net.www.protocol.https=ALL-UNNAMED</jvmFlag>
+            <jvmFlag>-Duser.timezone='GMT-5'</jvmFlag>
           </jvmFlags>
+
+
           <environment>
             <TEST_VAR>testVariableValue</TEST_VAR>
           </environment>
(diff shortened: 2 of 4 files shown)


src/test/resources/projects/standard-project/pom.xml
```diff
@@ -14,30 +14,79 @@
...
+              <version>1</version>
+              <project>travis-app-maven-plugin</project>
+              <version>GCLOUD_VERSION</version>
+              <jvmFlags>
+                <!-- required for kokoro gcp_ubuntu b/37217450 -->
+                <jvmFlag>-Djava.security.egd=file:/dev/urandom</jvmFlag>
+
+                <!-- Illegal reflective access disallowed since Java 16 -->
+                <jvmFlag>--add-opens</jvmFlag>
+                <jvmFlag>java.base/java.net=ALL-UNNAMED</jvmFlag>
```
mpeddada1, line 70: Would this work? 
```suggestion
                <jvmFlag>--add-opens=java.base/java.net=ALL-UNNAMED</jvmFlag>
```

src/test/resources/projects/standard-project/pom.xml
```diff
@@ -14,30 +14,79 @@
...
+            <version>${appengine-maven-plugin.version}</version>
+            <configuration>
+              <version>1</version>
+              <project>travis-app-maven-plugin</project>
+              <version>GCLOUD_VERSION</version>
+              <jvmFlags>
+                <!-- required for kokoro gcp_ubuntu b/37217450 -->
+                <jvmFlag>-Djava.security.egd=file:/dev/urandom</jvmFlag>
+
+                <!-- Illegal reflective access disallowed since Java 16 -->
```
mpeddada1, line 68: Out of curiosity, what do we get if we remove these `--add-opens` flags?

src/test/resources/projects/standard-project/pom.xml
```diff
@@ -14,30 +14,79 @@
...
-          </environment>
-          <project>travis-app-maven-plugin</project>
-        </configuration>
-      </plugin>
-    </plugins>
-  </build>
 
   <profiles>
+    <profile>
+      <id>java8_and_below</id>
```
mpeddada1, line 20: Would this count as a breaking change? Since we won't be able to run just `mvn package appengine:<goal>` (i.e without the profile) anymore.

mpeddada1:
So cool that you were able to get this to work!

src/test/resources/projects/standard-project/pom.xml
elefeint:
```
java.lang.RuntimeException: Unable to create a DevAppServer
	at com.google.appengine.tools.development.DevAppServerFactory.doCreateDevAppServer(DevAppServerFactory.java:378)
...
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make static java.net.URLStreamHandler java.net.URL.getURLStreamHandler(java.lang.String) accessible: module java.base does not "opens java.net" to unnamed module @73a28541
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
	at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
	at java.base/java.lang.reflect.Method.setAccessible(Method.java:193)
	at com.google.appengine.tools.development.StreamHandlerFactory.<clinit>(StreamHandlerFactory.java:52)

```

src/test/resources/projects/standard-project/pom.xml
elefeint: This is actually only a problem when running dev server. It's not related to the maven plugin at all.
This project has a `RunAsyncMojoIntegrationTest.testRunAsync()` test that starts the devserver and verifies that it starts successfully.

I have a standalone example here -- https://github.com/elefeint/examples/tree/master/appengine17-helloworld

elefeint: You are right that we should add this to the user guide, since `mvn appengine:run` will have this issue on Java 16 and up.

src/test/resources/projects/standard-project/pom.xml
elefeint: I thought it did not work, but it does!

src/test/resources/projects/standard-project/pom.xml
mpeddada1: Nice, I removed these flags and ran `mvn clean install` with java 17. It resulted in the following error:
```
: Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make static java.net.URLStreamHandler java.net.URL.getURLStreamHandler(java.lang.String) accessible: module java.base does not "opens java.net" to unnamed module @3ac3fd8b
testRunAsync: [INFO] GCLOUD: 	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
testRunAsync: [INFO] GCLOUD: 	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
testRunAsync: [INFO] GCLOUD: 	at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
testRunAsync: [INFO] GCLOUD: 	at java.base/java.lang.reflect.Method.setAccessible(Method.java:193)
testRunAsync: [INFO] GCLOUD: 	at com.google.appengine.tools.development.StreamHandlerFactory.<clinit>(StreamHandlerFactory.java:52)
testRunAsync: [INFO] GCLOUD: 	... 12 more
```

USER_GUIDE.md
```diff
@@ -39,6 +39,18 @@ For appengine-web.xml based projects the plugin exposes the following goals :
 | `start`           | Start the application in the background. |
 | `stop`            | Stop a running application. |
 
+NOTE: To run locally on Java 17, you'll need to add the following plugin configuration:
+
+```
+<configuration>
```
mpeddada1, line 45: Nice clarification!

src/test/resources/projects/standard-project/pom.xml
```diff
@@ -14,30 +14,76 @@
...
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.google.cloud.tools</groupId>
+            <artifactId>appengine-maven-plugin</artifactId>
+            <version>${appengine-maven-plugin.version}</version>
+            <configuration>
+              <version>1</version>
+              <project>travis-app-maven-plugin</project>
```
mpeddada1, line 32: Is it possible to change the project name here? I'm guessing it is specific to the CI.

mpeddada1 (approved):

src/test/resources/projects/standard-project/pom.xml
elefeint: It would be good to change at some point, since it likely is a throwback to when we were on Travis CI for many projects. But I did not introduce it in this PR, so I don't want to mess with it.

PR #464, opened 2022-04-14
workaround for permission

ddixit14:
last try: updating the path to /home/runner/work/app-maven-plugin/app-maven-plugin


zhumin8 (approved):
approving in case you still want to try this.

PR #466, opened 2022-04-14
release v2.4.2

chanseokoh:
To be merged after the release is complete.


chanseokoh:
The author of this auto-created PR is me because my personal token is being used. (This was the recommended way. Not sure if this is still the case though.) Not urgent at all, but you may want to [replace the token](https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/77c8ad7d24260719ca23b3062e7df83ad87748d6/.github/workflows/prepare-release.yml#L49-L53).

zhumin8 (approved):

zhumin8:
@chanseokoh Noted. Thanks for bringing this up!

